PR #663, opened 2015-01-02
reindex_everything should ignore non-RDF sources

jcoyne:
Because they don't have parsable graphs.

### DIFF
--- a/lib/active_fedora/indexing.rb
+++ b/lib/active_fedora/indexing.rb
@@ -79,6 +79,9 @@ module ActiveFedora
 
       def get_descendent_uris(uri)
         resource = Ldp::Resource::RdfSource.new(ActiveFedora.fedora.connection, uri)
+        # GET could be slow if it's a big resource, we're using HEAD to avoid this problem,
+        # but this causes more requests to Fedora.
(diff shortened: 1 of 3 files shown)


jcoyne:
@mjgiarlo this should solve the problem you have with https://gist.github.com/mjgiarlo/a15bc935ba26ab6723f2

mjgiarlo:
:+1:

mjgiarlo:
Comments from others?

acoburn:
Is it possible that you would have metadata attached to the non-RDF (binary) node? For instance, technical metadata or other preservation-related data specific to the binary (as opposed to the parent object). I at least, was planning to put certain types of technical metadata on the {binary path}/fcr:metadata endpoint. On the other hand, I probably won't be using solrizer for indexing, so don't consider this a particularly strong concern.

jcoyne:
@acoburn In that case the non-RDF's parent is responsible for doing the indexing of its children.  Regardless of that fact, this PR does not affect the current behavior in any way aside from preventing the stack trace that @mjgiarlo encountered.

acoburn:
@jcoyne thanks for the explanation!

Issue #653, opened 2014-12-17
Should be able to assign singular attribute to ActiveTriples::Resource


dchandekstark:
Current code will break b/c AT::Resource responds to :each.

Patch https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/attributes.rb#L99-L105 to account for this exception.

PR #676, opened 2015-01-07
Can assign single ActiveTriples::Resource to single-valued attribute (fi...

dchandekstark:
...xes #653)

Bumped version to 8.0.0.rc3

### DIFF
--- a/lib/active_fedora/attributes.rb
+++ b/lib/active_fedora/attributes.rb
@@ -83,11 +83,13 @@ module ActiveFedora
       end
       raise UnknownAttributeError, "#{self.class} does not have an attribute `#{field}'" unless self.class.defined_attributes.key?(field)
       if self.class.multiple?(field)
-        unless args.nil? || args.respond_to?(:each)
-          raise ArgumentError, "Cannot set the multi-valued attribute `#{field}' to a scalar value."
+        unless args.nil? || ( args.respond_to?(:each) && !args.kind_of?(ActiveTriples::Resource) )
+          raise ArgumentError, "Cannot set the multi-valued attribute `#{field}' to a single value."
         end
       elsif args.respond_to?(:each) # unique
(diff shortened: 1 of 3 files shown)


dchandekstark:
If this passes Travis and gets :+1: 'd I'd propose making this 8.0.0 final.

dchandekstark:
Ping @jcoyne @terrellt @mjgiarlo

terrellt:
@dchandekstark :+1: I assume this works like AT, in that if I set a property to an AT::Resource it sets it to a URI which spins up a resource on access.

dchandekstark:
@terrellt No change in how the property setting works per se, this just fixes #653 which would not permit setting a single-valued AF model attribute to an AT Resource b/c an AT::Resource is an RDF::Graph, which is an enumerable.

mjgiarlo:
:ok_hand: by me.

terrellt:
Are there tests already ensuring that you can set a property to an AT::Resource and get it back out?

dchandekstark:
@terrellt Dunno, but that would be a separate issue, right?

terrellt:
> @terrellt Dunno, but that would be a separate issue, right?

Maybe? This PR specifically allows something that could be broken. If that's okay with the community, then I yield to that judgement. @mjgiarlo @jcoyne ?

dchandekstark:
@terrellt Not exactly.  What's being allowed here was allowed in version 7 (with a deprecation warning).  In Version 8 it breaks because you can't set the value of single-valued attribute to something that responds to :each (prior to this PR).  That condition was written with Array in mind -- it wasn't intended to apply to an AT::Resource.  If there's a deeper issue here, I need more context.  It's possible that something *is* broken, although I don't see this PR as doing the breaking.

jcoyne:
:+1:

Issue #662, opened 2014-12-22
There should be a warning if you specify two properties that share the same predicate.


jcoyne:
Would help avoid bugs like this:
```
     Failure/Error: let(:object) { TestModel.new(title: ['foo', 'bar'], creator: 'baz') }
     ActiveFedora::ConstraintError:
       Expected "creator" to have 0-1 statements, but there are 2
     # ./spec/presenters/hydra_editor_presenter_spec.rb:23:in `new'
```

Which was caused by title and creator accidentally having the same predicate

jcoyne:
A predicate can be shared if each term has different `class_name` properties. So the warning should only be raised if the cardinality (`multiple` property) is different on both.

hectorcorrea:
@jcoyne A few questions. 

1. Why does the "class_name" matter?

2. Is this the expected behavior for when two properties have the same predicate and 
 
a) if the have the same "class_name" and different "multiple" - OK
b) if the have the same "class_name" and same "multiple" - Warn user
c) if the have the different "class_name" and different "multiple" - OK
d) if the have the different "class_name" and same "multiple" - OK

3. If one (or both) of them have no "class_name" indicated apply the same logic?

jcoyne:
Actually I think the class_name difference only applies to associations (belongs_to, has_many). I may have confused that in my head, so it should probably always warn on duplicate predicates regardless of how any other parameters are set.

hectorcorrea:
@jcoyne I have this note form our IRC chat the other day on which @terrellt brought up "class_name" 

    AT can handle property 

    :title,   :class_name => Title1, :predicate => RDF::DC.title, 
    :title_2, :class_name => Title2, :predicate => RDF::DC.title

    if Title1 and Title2 have defined types which differ.

Is this why you included it perhaps? I've removed the "class_name" check in my PR but I wonder if this changes things.

jcoyne:
Perhaps. It's not a use case I have, see what @terrellt has to say.

PR #681, opened 2015-01-08
Warn when the same predicate is used in more than one property

hectorcorrea:
This addresses issue #662

### DIFF
--- a/lib/active_fedora/attributes.rb
+++ b/lib/active_fedora/attributes.rb
@@ -171,15 +171,22 @@ module ActiveFedora
       end
 
       def property name, properties={}, &block
+        warn_duplicate_predicates name, properties
         properties = { multiple: true }.merge(properties)
         find_or_create_defined_attribute(name, nil, properties)
         raise ArgumentError, "#{name} is a keyword and not an acceptable property name." if protected_property_name? name
         reflection = ActiveFedora::Attributes::PropertyBuilder.build(self, name, properties, &block)
-        # reflection = ActiveTriple::PropertyBuilder.build(self, name, properties, &block)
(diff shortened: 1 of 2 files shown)


lib/active_fedora/attributes.rb
```diff
@@ -171,16 +171,37 @@ def multiple?(field)
...
         ActiveTriples::Reflection.add_reflection self, name, reflection
       end
 
       private
 
+      def warn_duplicate_predicates new_name, new_properties
+        # TODO: Simplify logic once I am sure I got it right.
+        self.properties.each do |property, config| 
+          if config.predicate == new_properties[:predicate]
+            puts "- potential issue for predicate #{config.predicate} between #{new_name} and #{property}"
```
jcoyne: This should be `ActiveFedora::Base.logger`

lib/active_fedora/attributes.rb
```diff
@@ -171,16 +171,37 @@ def multiple?(field)
...
+            if config.class_name == new_properties[:class_name]
+              if config.multiple? == new_properties[:multiple]
+                puts "\twe have a problem, warn the user"
+                # ActiveFedora::Base.logger.warn "Same predicate (#{config.predicate}) used for properties #{new_name} and #{property}"
+                ActiveFedora::Base.logger.warn "Same predicate"
+              else
+                puts "\tno real issue (different cardinality)"
```
jcoyne: I think this is one of the issues we're trying to avoid actually.

hectorcorrea:
@jcoyne Updated to warn based on predicate alone (as per https://github.com/projecthydra/active_fedora/issues/662#issuecomment-69339032) Let me know what you think.

jcoyne:
:+1:

Issue #656, opened 2014-12-18
Misleading 302 error from LDP when pointing to wrong Fedora URL


flyingzumwalt:
If fedora.yml points to /fedora instead of /fedora/rest, you get `Ldp::HttpError: STATUS: 302 ...` and an unhelpful backtrace.  

1. The 302 is misleading
2. A lot of people are going to run into this when they upgrade.  It should be handled with informative feedback (ie. check the fedora config for "/rest" and suggest that might be the problem if the url does not include it.

hectorcorrea:
Fixed on PR #684

Issue #680, opened 2015-01-08
If base_path isn't set then Object.URI is wrong.


jcoyne:
If I don't have base_path set in fedora.yml
then when I create an object: `o = ActiveFedora::Base.find('5831ea7c-64c2-4870-8001-8d1ce533ebce')`
then its uri has a blank path segment (`//`) in it. i.e. 
"http://127.0.0.1:8080/fedora/rest//58/31/ea/7c/5831ea7c-64c2-4870-8001-8d1ce533ebce"

The graph returned from accessing that path, doesn't have the blank path segment, so the assertions are not returned.

jcoyne:
To elaborate, there is no reason a user must configure 'base_path', it's just a convenience for scoping your current connection to a specific subset of URIs

hectorcorrea:
Duplicate of issue #657 
Fixed in PR #696

PR #684, opened 2015-01-13
Warn user when initial connection to Fedora fails and the URL does not e...

hectorcorrea:
Addresses issue https://github.com/projecthydra/active_fedora/issues/656

### DIFF
--- a/lib/active_fedora/fedora.rb
+++ b/lib/active_fedora/fedora.rb
@@ -26,6 +26,9 @@ module ActiveFedora
         ActiveFedora::Base.logger.info "Attempted to init base path `#{root_resource_path}`, but it already exists" if ActiveFedora::Base.logger
         return false
     rescue Ldp::NotFound
+      if !host.downcase.end_with?("/rest")
+        ActiveFedora::Base.logger.warn "Fedora URL (#{host}) does not end with /rest. This could be a problem. Check your fedora.yml config"
+      end
       connection.put(root_resource_path, BLANK).success?
     end
 
--- a/spec/unit/active_fedora_spec.rb
+++ b/spec/unit/active_fedora_spec.rb
@@ -15,6 +15,24 @@ describe ActiveFedora do
     restore_spec_configuration
   end
 
+  describe "validate Fedora URL" do 
+
+    let(:good_url) { ActiveFedora.fedora_config.credentials[:url] }
+    let(:bad_url) { good_url.gsub('/rest', '/') }
+
+    it "should connect OK" do 
+      expect(ActiveFedora::Base.logger).to_not receive(:warn)
+      ActiveFedora::Fedora.new(url: good_url, base_path: '/test')
+    end
+
+    it "should not connect and warn" do 
+      expect(ActiveFedora::Base.logger).to receive(:warn)
+      expect {
+        ActiveFedora::Fedora.new(url: bad_url, base_path: '/test')
+      }.to raise_error Ldp::HttpError 
+    end
+  end
+
   describe "initialization methods" do
     describe "environment" do
       it "should use config_options[:environment] if set" do


lib/active_fedora/fedora.rb
```diff
@@ -26,6 +26,9 @@ def init_base_path
         ActiveFedora::Base.logger.info "Attempted to init base path `#{root_resource_path}`, but it already exists" if ActiveFedora::Base.logger
         return false
     rescue Ldp::NotFound
+      if !host.downcase.end_with?("/rest")
+        ActiveFedora::Base.logger.warn "Fedora URL (#{host}) does not end with /rest. This could be the problem. Check your fedora.yml config"
```
jcoyne: Switch "This could be the problem" to "This could be a problem"

jcoyne:
This assumes they will never modify https://github.com/fcrepo4/fcrepo4/blob/3bfc11f2a3088f7435e47268faafb4f000ae50c0/fcrepo-webapp/src/main/webapp/WEB-INF/web.xml#L33 is that a safe assumption?

spec/unit/active_fedora_spec.rb
```diff
@@ -15,6 +15,24 @@
...
   end
 
+  describe "validate Fedora URL" do 
+
+    let(:good_url) { ActiveFedora.fedora_config.credentials[:url] }
+    let(:bad_url) { good_url.gsub('/rest', '/') }
+
+    it "should connect OK" do 
+      expect(ActiveFedora::Base.logger).to_not receive(:warn)
+      ActiveFedora::Fedora.new(:url => good_url, :base_path => '/test')
```
mjgiarlo: Use Ruby 1.9-style hashes, pls, good sir. (Kill the hashrockets!)

spec/unit/active_fedora_spec.rb
```diff
@@ -15,6 +15,24 @@
...
+    end
+
+    it "should not connect and warn" do 
+      expect {
+        expect(ActiveFedora::Base.logger).to receive(:warn)
```
mjgiarlo: Should this line be on the previous line, outside the `expect { ... }.to raise_error...`?

spec/unit/active_fedora_spec.rb
```diff
@@ -15,6 +15,24 @@
...
+        ActiveFedora::Fedora.new(:url => bad_url, :base_path => '/test')
```
mjgiarlo: Use Ruby 1.9-style hashes, pls, good sir. (Kill the hashrockets!)

jcoyne:
Another way to do this would be to get the resource at `host` and see if it is returning an RDF graph.

mjgiarlo:
@jcoyne I'm not too concerned about the web.xml thing, because the warning only appears when an LDP error is raised, and the message is merely a suggestion about what might be an easy mistake to make for folks new to F4.

Though I wonder if another quick check might be taking a glance at the base_path value and making sure it exists?  That's tripped one or two of us up already.

jcoyne:
@mjgiarlo this ticket addresses #656, you are bringing up issue #680

spec/unit/active_fedora_spec.rb
hectorcorrea: @mjgiarlo They are now gone [insert sad :rocket: here]

spec/unit/active_fedora_spec.rb
hectorcorrea: @mjgiarlo They are now gone [insert sad :rocket: here]

spec/unit/active_fedora_spec.rb
hectorcorrea: @mjgiarlo It worked both ways, but I moved it out as you suggested to prevent confusion.

hectorcorrea:
@mjgiarlo @jcoyne I am not to worried about users tweaking the web.xml because as Mike pointed out the warning is only raised if the connection fails. I suspect users that intentionally update the Fedora settings will know what to make out of the warning.

mjgiarlo:
w00t, :rocket: is :bomb:

flyingzumwalt:
:+1:  :wink:

PR #689, opened 2015-01-14
Pass nested attribute options to the resource

jcoyne:


mjgiarlo:
:+1: -- and where have we documented what the possible attribute options are?

jcoyne:
@mjgiarlo https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/nested_attributes.rb#L22-L46

PR #691, opened 2015-01-16
Use Fcrepo4 class repository definitions

awead:
Use Fcrepo4 class throughout.

### DIFF
--- a/lib/active_fedora/file.rb
+++ b/lib/active_fedora/file.rb
@@ -118,7 +118,7 @@ module ActiveFedora
     end
 
     def digest
-      response = metadata.ldp_source.graph.query(:predicate => ::RDF::URI.new("http://fedora.info/definitions/v4/repository#digest"))
+      response = metadata.ldp_source.graph.query(predicate: ActiveFedora::RDF::Fcrepo4.digest)
       response.map(&:object)
     end
 
--- a/lib/active_fedora/fixity_service.rb
+++ b/lib/active_fedora/fixity_service.rb
@@ -18,7 +18,7 @@ module ActiveFedora
     end
 
     def status
-      fixity_graph.query(predicate: status_url).map(&:object).first.to_s
+      fixity_graph.query(predicate: ActiveFedora::RDF::Fcrepo4.status).map(&:object).first.to_s
     end
 
     private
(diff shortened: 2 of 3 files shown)


lib/active_fedora/fixity_service.rb
```diff
@@ -32,10 +32,7 @@ def fixity_graph
       ::RDF::Graph.new << ::RDF::Reader.for(:ttl).new(response.body)
     end
 
-    def status_url
-      ::RDF::URI("http://fedora.info/definitions/v4/repository#status")
```
mjgiarlo: Why not leave this method and make it return `ActiveFedora::RDF::Fcrepo4.status`?  That way it's more easily overridable.

lib/active_fedora/file.rb
```diff
@@ -118,7 +118,7 @@ def original_name
     end
 
     def digest
-      response = metadata.ldp_source.graph.query(:predicate => ::RDF::URI.new("http://fedora.info/definitions/v4/repository#digest"))
+      response = metadata.ldp_source.graph.query(predicate: ActiveFedora::RDF::Fcrepo4.digest)
```
mjgiarlo: Similar to  my comment on the FixityService, how about adding a new method to this class called `digest_url` and have it return `ActiveFedora::RDF::Fcrepo4.digest` so that folks can more easily override this downstream with a URI of their choosing?

mjgiarlo:
Requesting a couple of small changes that allow downstream adopters to more easily override URIs. If you're :-1: on those changes, I can be flexible.

lib/active_fedora/fixity_service.rb
jcoyne: I can't see a case where you'd want to override this. Fcrepo will always use this predicate. I can't see this code being repurposed for anything else.

jcoyne:
:+1: as it is

lib/active_fedora/fixity_service.rb
mjgiarlo: :ok_hand:

mjgiarlo:
:+1: as it is as well. Merge, @jcoyne?

PR #694, opened 2015-01-20
Load singular datastream attributes from solr

jcoyne:
Previously only the first character of singular datastream attributes
was being returned when the object was loaded from solr.


mjgiarlo:
:+1: if Travis is happy.

Issue #657, opened 2014-12-19
If base_path is not set in fedora.yml, auto-casting doesn't work in ActiveFedora::Base.find


flyingzumwalt:
If your fedora.yml looks like this:
```
development:
  user: fedoraAdmin
  password: fedoraAdmin
  url: http://127.0.0.1:8983/fedora/rest
```
instead of this:
```
development:
  user: fedoraAdmin
  password: fedoraAdmin
  url: http://127.0.0.1:8983/fedora/rest
  base_path: ""
```

Then you get this behavior:
```ruby
book = MyBookModel.create(title:["foo"])
ActiveFedora::Base.find(book.id)
=> ActiveFedora::Base(id:"xyz")
```
instead of the desired:
```ruby
book = MyBookModel.create(title:["foo"])
ActiveFedora::Base.find(book.id)
=> MyBookModel(id:"xyz")
```

hectorcorrea:
Fixed in PR #696

PR #696, opened 2015-01-22
Fixes bug that prevented ActiveFedora from deserializing classes 

hectorcorrea:
correctly when the base_path setting in fedora.yaml was not set.

Takes care of https://github.com/projecthydra/active_fedora/issues/657

Note: When I removed the base_path from config/fedora.yml and ran all tests I got 20+ failing tests. I spot checked several of them and the issues where because the tests were hard-coded to expect a /test/ on the URL or because the cleaner didn't know how to clean from fedora/rest rather than fedora/rest/test. I didn't think this should be an issue with the code given the new tests that evaluate the new changes.

### DIFF
--- a/lib/active_fedora/core.rb
+++ b/lib/active_fedora/core.rb
@@ -116,6 +116,9 @@ module ActiveFedora
       end
 
     module ClassMethods
+
+      SLASH = '/'.freeze
+
       def generated_association_methods
         @generated_association_methods ||= begin
           mod = const_set(:GeneratedAssociationMethods, Module.new)
@@ -138,8 +141,10 @@ module ActiveFedora
         if translate_id_to_uri
           translate_id_to_uri.call(id)
         else
-          id = "/#{id}" unless id.start_with? '/'
-          id = ActiveFedora.fedora.base_path + id unless id.start_with? "#{ActiveFedora.fedora.base_path}/"
+          id = "/#{id}" unless id.start_with? SLASH
+          unless ActiveFedora.fedora.base_path == SLASH || id.start_with?("#{ActiveFedora.fedora.base_path}/")
+            id = ActiveFedora.fedora.base_path + id 
+          end
           ActiveFedora.fedora.host + id
         end
       end
--- a/spec/unit/core_spec.rb
+++ b/spec/unit/core_spec.rb
@@ -104,6 +104,20 @@ describe ActiveFedora::Base do
 
       it { should eq "#{ActiveFedora.fedora.host}#{ActiveFedora.fedora.base_path}/foo/123456w" }
     end
+
+    context "with an empty base path" do
+      it "should produce a valid URI" do
+        allow(ActiveFedora.fedora).to receive(:base_path).and_return("/")    
+        expect(subject).to eq("#{ActiveFedora.fedora.host}/#{id}")
+      end
+    end
+
+    context "with a really empty base path" do
+      it "should produce a valid URI" do
+        allow(ActiveFedora.fedora).to receive(:base_path).and_return("")    
+        expect(subject).to eq("#{ActiveFedora.fedora.host}/#{id}")
+      end
+    end
   end
 
   describe "uri_to_id" do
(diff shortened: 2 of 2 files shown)


lib/active_fedora/core.rb
```diff
@@ -139,7 +139,9 @@ def id_to_uri(id)
           translate_id_to_uri.call(id)
         else
           id = "/#{id}" unless id.start_with? '/'
-          id = ActiveFedora.fedora.base_path + id unless id.start_with? "#{ActiveFedora.fedora.base_path}/"
+          unless ActiveFedora.fedora.base_path == '/'
```
jcoyne: Would you please combine the two conditionals: 
`unless ActiveFedora.fedora.base_path == '/' || id.start_with? "#{ActiveFedora.fedora.base_path}/"`

also freeze the string to prevent allocation on each trip through here: `'/'.freeze`

spec/unit/core_spec.rb
```diff
@@ -123,4 +123,28 @@ class Book < ActiveFedora::Base
       it { should eq '123456w' }
     end
   end
+
+  describe "with an empty base_path" do 
+    let(:book) { Book.create(title:"foo") }
```
jcoyne: This can be factored out into a describe block that contains all three of the blocks you added. Also factor out:
```
allow(ActiveFedora.fedora).to receive(:base_path).and_return(base)
```
Each block can `let(:base) { '/' }` as appropriate

jcoyne: I'm wondering if it wouldn't be more direct to just test `id_to_uri` as that's the behavior you've changed.

lib/active_fedora/core.rb
hectorcorrea: @jcoyne done

spec/unit/core_spec.rb
hectorcorrea: @jcoyne I updated the tests to remove the duplicate code. I am torn between leaving them as they are now (validating the casting to the right class) vs validating the URL. I can go either way.

spec/unit/core_spec.rb
```diff
@@ -123,4 +123,22 @@ class Book < ActiveFedora::Base
       it { should eq '123456w' }
     end
   end
+
+  describe "when serializing an object" do 
+    let(:book) { Book.create(title:"foo") }
+    it "should handle an empty base path" do
+      allow(ActiveFedora.fedora).to receive(:base_path).and_return("/")
+      expect(ActiveFedora::Base.find(book.id).class.name).to eq("Book")
```
jcoyne: How about testing `id_to_uri` directly rather than doing create/lookup to test that change?

hectorcorrea: OK. I'll make that change tomorrow and resubmit.

jcoyne: If you want to validate that the correct class is found, you should move the tests to spec/integration, because that would be an integration test not a unit test.

hectorcorrea: OK. I've updated the tests to go directly against ActiveFedora::Base.id_to_uri and bypass Fedora.

spec/unit/core_spec.rb
```diff
@@ -104,6 +104,26 @@ class Book < ActiveFedora::Base
...
+    context "with a really empty base path" do
+      it "should produce a valid URI" do
+        allow(ActiveFedora.fedora).to receive(:base_path).and_return("")    
+        expect(subject).to eq("#{ActiveFedora.fedora.host}/#{id}")
+      end
+    end
+
+    context "with the default base path" do
+      it "should produce a valid URI with the base path on it" do
+        expect(subject).to eq("#{ActiveFedora.fedora.host}#{ActiveFedora.fedora.base_path}/#{id}")
```
jcoyne: This test is now a duplicate of what is on line 96. Can we just remove it?

hectorcorrea: It's gone now.

Issue #702, opened 2015-01-26
Uniform application of dates


awead:
Any ldp resource from Fedora has a `repository#created` and `repository#lastModified` property. We should be able to return that value whether it is a file, object, metadata path, or version, and use consistent method names. For example, `create_date` should return _created_ from the ldp source, and `modified_date` should return the _lastModified_ from the ldp source.

Returning `last_modified` on a file appears to be here:
https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/file.rb#L15

But that isn't doing what we think it is. You'd have to call `created` on the most recent version of a file to get this value,

PR #703, opened 2015-01-26
File#last_modified removed

jcoyne:


mjgiarlo:
Can you say more about why this change is needed?

awead:
@mjgiarlo this was old code that I discovered. `last_modified` isn't doing anything. It'll always return nil because it's never getting set. I've explained what should happen to rectify this situation in #702

mjgiarlo:
OK, just wanted a record of it on GitHub.

PR #707, opened 2015-01-28
File#stream should return a FileBody object

jcoyne:
instead of yielding directly.
Attempting to follow this pattern:
https://github.com/rails/rails/blob/08754f12e65a9ec79633a605e986d0f1ffa4b251/actionpack/lib/action_controller/metal/data_streaming.rb#L81-L96
which should let us set `self.response_body = file.stream` in a
controller


mjgiarlo:
:100:

PR #711, opened 2015-02-08
Avoid unnecessary HEAD request on file retrieval. Fixes #493

### LINKED ISSUE #493: Opitimize HTTP interaction with Fedora 4
Are we doing unnecessary HEAD requests?

jcoyne:
If a NonRDFSource has a parent which is an RDFSource (AF::Base) and the
parent indicates that it contains the file (ldp:contains) then there is
no need to send a HEAD request to see if the file exists.


mjgiarlo:
Major fails there.

mjgiarlo:
Looks like it's related to a missing dependency on 'zip' from jettywrapper. @barmintor :interrobang: :interrobang: :interrobang:

PR #715, opened 2015-02-17
Provide more documentation around delete and destroy [ci skip]

jcoyne:


acozine:
This is helpful - thank you.

PR #719, opened 2015-02-24
Jettywrapper doesn't require hydra_jetty_version

awead:
As of Jettywrapper 2.0.3, the hydra_jetty_version is no longer required unless you want to download a version of hydra-jetty other than master.


mjgiarlo:
:+1:

PR #720, opened 2015-02-28
Removing deprication warning about Rsolr.escape

cam156:


jcoyne:
Better update the rsolr version in the gemspec

cam156:
Well that did not work...

PR #722, opened 2015-02-28
Use modified RSolr.solr_escape method

awead:
Merge this depending on what you're thoughts are on rsolr/rsolr#101

### DIFF
--- a/lib/active_fedora/relation/finder_methods.rb
+++ b/lib/active_fedora/relation/finder_methods.rb
@@ -264,10 +264,10 @@ module ActiveFedora
         if value.empty?
           "-#{key}:['' TO *]"
         elsif value.is_a? Array
-          value.map { |val| "#{key}:#{RSolr.escape(val)}" }
+          value.map { |val| "#{key}:#{SolrQueryBuilder.solr_escape(val)}" }
         else
           key = SOLR_DOCUMENT_ID if (key === :id || key === :id)
-          "#{key}:#{RSolr.escape(value)}"
+          "#{key}:#{SolrQueryBuilder.solr_escape(value)}"
         end
       end
     end
--- a/lib/active_fedora/solr_query_builder.rb
+++ b/lib/active_fedora/solr_query_builder.rb
@@ -35,6 +35,12 @@ module ActiveFedora
       clauses.empty? ? "id:NEVER_USE_THIS_ID" : clauses.join(" #{join_with} ".freeze)
     end
 
+    # Adds esaping for spaces which are not handled by RSolr.solr_escape
+    # See rsolr/rsolr#101
+    def self.solr_escape terms
+      RSolr.solr_escape(terms).gsub(/\s+/,"\\ ")
+    end
+
     private
       # Given an list of 2 element lists, transform to a list of solr clauses
       def self.pairs_to_clauses(pairs)
--- a/spec/integration/query_result_builder_spec.rb
+++ b/spec/integration/query_result_builder_spec.rb
@@ -30,7 +30,7 @@ describe ActiveFedora::QueryResultBuilder do
       Object.send(:remove_const, :FooObject)
     end
     it "should return an array of objects that are of the class stored in active_fedora_model_s" do
-      query = "id\:#{RSolr.escape(@test_object.id)} OR id\:#{RSolr.escape(@foo_object.id)}"
+      query = ActiveFedora::SolrQueryBuilder.construct_query_for_ids([@test_object.id, @foo_object.id])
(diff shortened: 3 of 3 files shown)


lib/active_fedora/solr_query_builder.rb
```diff
@@ -35,6 +35,12 @@ def self.construct_query_for_rel(field_pairs, join_with = 'AND')
       clauses.empty? ? "id:NEVER_USE_THIS_ID" : clauses.join(" #{join_with} ".freeze)
     end
 
+    # Adds esaping for spaces which are not handled by RSolr.solr_escape
+    # See rsolr/rsolr#101
+    def self.solr_escape terms
+      RSolr.solr_escape(terms).gsub(/\s+/,"\\ ")
+    end
```
cam156: @awead Should this be a change to rsolr instead of here?

jcoyne: @cam156 I don't think so. RSolr was escaping spaces before, but solr doesn't *need* spaces escaped. The escaped spaces just gives the query a different meaning.

spec/integration/query_result_builder_spec.rb
```diff
@@ -30,7 +30,7 @@ def self.id_namespace
       Object.send(:remove_const, :FooObject)
     end
     it "should return an array of objects that are of the class stored in active_fedora_model_s" do
-      query = "id\:#{RSolr.escape(@test_object.id)} OR id\:#{RSolr.escape(@foo_object.id)}"
+      query = "id\:#{ActiveFedora::SolrQueryBuilder.solr_escape(@test_object.id)} OR id\:#{ActiveFedora::SolrQueryBuilder.solr_escape(@foo_object.id)}"
```
jcoyne: The query probably should be `SolrQueryBuilder.construct_query_for_ids([@test_object.id, @foo_object.id])`

lib/active_fedora/relation/finder_methods.rb
```diff
@@ -264,10 +264,10 @@ def condition_to_clauses(key, value)
         if value.empty?
           "-#{key}:['' TO *]"
         elsif value.is_a? Array
-          value.map { |val| "#{key}:#{RSolr.escape(val)}" }
+          value.map { |val| "#{key}:#{SolrQueryBuilder.solr_escape(val)}" }
         else
           key = SOLR_DOCUMENT_ID if (key === :id || key === :id)
-          "#{key}:#{RSolr.escape(value)}"
+          "#{key}:#{SolrQueryBuilder.solr_escape(value)}"
```
jcoyne: This should be `SolrQueryBuilder.construct_query_for_ids(value)`

spec/integration/query_result_builder_spec.rb
```diff
@@ -40,7 +40,7 @@ def self.id_namespace
     end
 
     it 'should #reify a lightweight object as a new instance' do
-      query = "id\:#{RSolr.escape(@foo_object.id)}"
+      query = "id\:#{ActiveFedora::SolrQueryBuilder.solr_escape(@foo_object.id)}"
```
jcoyne: This should be `SolrQueryBuilder.construct_query_for_ids`

lib/active_fedora/relation/finder_methods.rb
```diff
@@ -264,10 +264,10 @@ def condition_to_clauses(key, value)
         if value.empty?
           "-#{key}:['' TO *]"
         elsif value.is_a? Array
-          value.map { |val| "#{key}:#{RSolr.escape(val)}" }
+          value.map { |val| "#{key}:#{SolrQueryBuilder.solr_escape(val)}" }
```
jcoyne: It seems that this is the only place that solr_escape is needed (after making my other suggested changes). Can we make it private?

spec/integration/query_result_builder_spec.rb
awead: @jcoyne but if you do that, are we loosing the "OR" syntax? I forget which Solr defaults to, but it can be either "AND" or "OR" depending on how it's configured.

awead: @jcoyne nm. I see now.

lib/active_fedora/relation/finder_methods.rb
awead: This is causing other errors. What if key isn't `:id` ?

jcoyne: Ah, I read the conditional on 269 wrong. I though this whole block was for ID.

awead:
@jcoyne let me know if the latest commit addresses all your suggestions.

PR #729, opened 2015-03-17
Allow a has_many association to specify an explicit foreign key via the :as option

mbklein:
Fixes the bug demonstrated in [this gist](https://gist.github.com/mbklein/88d07b3fe6ad97ee5f1d), which I never got around to filing an issue about because I fixed it instead.

### DIFF
--- a/spec/integration/associations_spec.rb
+++ b/spec/integration/associations_spec.rb
@@ -24,6 +24,37 @@ describe ActiveFedora::Base do
     end
 
   end
+  
+  describe "explicit foreign key" do
+    before do
+      class FooThing < ActiveFedora::Base
+        has_many :bars, :class_name=>'BarThing', :predicate=>ActiveFedora::RDF::Fcrepo::RelsExt.isPartOf, :as=>:foothing
+      end
+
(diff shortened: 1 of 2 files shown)


spec/integration/associations_spec.rb
```diff
@@ -24,6 +24,37 @@ class Person < ActiveFedora::Base
     end
 
   end
+  
+  describe "explicit foreign key" do
+    before do
+      class FooThing < ActiveFedora::Base
+        has_many :bars, :class_name=>'BarThing', :predicate=>ActiveFedora::RDF::Fcrepo::RelsExt.isPartOf, :as=>:foothing
```
awead: Do we want to make "foothing" implicit, but overrideable? If the class was named "Page" then it seems implicit that `:as => :page`.

mbklein: This is testing the override. The implicit key was `foo_thing`

awead: :ok:

mbklein: :cool:

PR #730, opened 2015-03-18
Make sure datastreams get configured on load as well as new/create

mbklein:
This PR fixes an edge-case bug described in [@cjcolvar's gist](https://gist.github.com/cjcolvar/86f8c87f7be86c08fa62), in which loading a resource with a SimpleDatastream will fail if the resource's class hasn't been configured with a call to `#new` or `#create` yet.

The corresponding test has to set up the fixture via `Ldp::Orm` in order to prevent the test class from being configured during setup. Hence the ugly test. If there's a way to make it prettier, I'd welcome some feedback.

### DIFF
--- a/spec/unit/simple_datastream_spec.rb
+++ b/spec/unit/simple_datastream_spec.rb
@@ -62,5 +62,37 @@ describe ActiveFedora::SimpleDatastream do
       expect(solr[ActiveFedora::SolrQueryBuilder.solr_name('creation_date', type: :date)]).to eq "2012-01-15"
     end
   end
+  
+  describe "datastream configuration" do
+    let(:foo) do
+      ActiveFedora::Base.create! do |obj|
+        obj.add_file(%{<?xml version="1.0"?>\n<fields><fubar>test</fubar></fields>}, path:'someData')
+      end
+    end
+    let(:resource) { Ldp::Resource::RdfSource.new(ActiveFedora.fedora.connection, foo.uri) }
+    let(:orm) { Ldp::Orm.new(resource) }
+
+    before do
+      class FooHistory < ActiveFedora::Base
+        has_metadata :type=>ActiveFedora::SimpleDatastream, :name=>"someData" do |m|
+          m.field "fubar", :string
+        end
+        has_attributes :fubar, datastream: 'someData', multiple: false
+      end
+
+      orm.graph.delete([orm.resource.subject_uri, ActiveFedora::RDF::Fcrepo::Model.hasModel, nil])
+      orm.graph.insert([orm.resource.subject_uri, ActiveFedora::RDF::Fcrepo::Model.hasModel, 'FooHistory'])
+      orm.save
+      foo.reload
+      foo.update_index
+    end
+    
+    after do
+      Object.send(:remove_const, :FooHistory)
+    end
+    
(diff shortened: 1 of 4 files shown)


spec/unit/simple_datastream_spec.rb
```diff
@@ -62,5 +62,41 @@
...
+      f.reload
+      f.update_index
+      f.id
+    }
+    
+    let(:foo) { FooHistory.find(foo_id) }
+    
+    it "should configure the datastream on load" do
+      expect(foo.fubar).to eq 'test'
+    end
```
awead: The above 5 lines could be simplified to:
```
subject { FooHistory.find(foo_id).fubar }
it { is_expected.to eq "test" }
```
I don't have any ideas about the big `let` block, though.

mjgiarlo: Or, using `rspec-its`:

```ruby
subject { FooHistory.find(foo_id) }
its(:fubar) { is_expected.to eq 'test' }
```

spec/unit/simple_datastream_spec.rb
```diff
@@ -62,5 +62,41 @@
...
+      client = ActiveFedora.fedora.connection
+      resource = Ldp::Resource::RdfSource.new(client, f.uri)
+      orm = Ldp::Orm.new(resource)
+      orm.graph.delete([orm.resource.subject_uri, ActiveFedora::RDF::Fcrepo::Model.hasModel, nil])
+      orm.graph.insert([orm.resource.subject_uri, ActiveFedora::RDF::Fcrepo::Model.hasModel, 'FooHistory'])
+      orm.save
+      f.reload
+      f.update_index
+      f.id
+    }
```
mjgiarlo: I might rework this like so:

```ruby
let!(:foo) do
  f = ActiveFedora::Base.create! do |job|
    obj.add_file(...)
  end
end
let(:resource) { Ldp::Resource::RdfSource.new(ActiveFedora.fedora.connection, foo.uri) }
let(:orm) { Ldp::Orm.new(resource) }

before do
  orm.graph.delete([orm.resource.subject_uri, ActiveFedora::RDF::Fcrepo::Model.hasModel, nil])
  orm.graph.insert([orm.resource.subject_uri, ActiveFedora::RDF::Fcrepo::Model.hasModel, 'FooHistory'])
  orm.save
  foo.reload
  foo.update_index
end

subject { FooHistory.find(foo.id) }
its(:fubar) { is_expected.to eq 'test' }
```

awead: :+1:

mjgiarlo: Principles I followed here, just to be clear where I'm coming from, are:

1. Pass around objects rather than values (`id`) from within objects.
2. Declare variables used elsewhere in `let` statements.
3. Side effects or preconditions go in the `before` block.

awead:
@mbklein would you mind squashing these? I think one commit is fine, unless you wanted to have the separation.

mbklein:
:ok:

mjgiarlo:
:clap:

mbklein:
:shipit:

mjgiarlo:
@awead :fist:

cjcolvar:
:clap: Thanks @mbklein for finishing this!

PR #735, opened 2015-03-25
Properties named *_id should not break the change set

cjcolvar:


lib/active_fedora/change_set.rb
```diff
@@ -19,7 +19,7 @@ def empty?
 
     def changes
       @changes ||= changed_attributes.each_with_object({}) do |key, result|
-        if m = /^.*_ids?$/.match(key)
+        if (m = /^.*_ids?$/.match(key)) && object.association(m[0].to_sym).present?
```
jcoyne: Is the first clause necessary any longer? Did you try changing this line to just have: `object.association(m[0].to_sym).present?`

lib/active_fedora/change_set.rb
```diff
@@ -19,8 +19,8 @@ def empty?
 
     def changes
       @changes ||= changed_attributes.each_with_object({}) do |key, result|
-        if m = /^.*_ids?$/.match(key)
-          predicate = object.association(m[0].to_sym).reflection.predicate
+        if object.respond_to?(:association) && object.association(key.to_sym).present?
```
jcoyne: why are you testing to see if object `responds_to?(:association)`? Isn't that always true?

cjcolvar: There was a test failing when changes was being called on a File which doesn't respond to association.

https://github.com/projecthydra/active_fedora/blob/master/spec/integration/with_metadata_spec.rb#L43

jcoyne: Okay, I think that's because fcr:metadata nodes don't have associations.

PR #743, opened 2015-03-27
Move the indexing logic to the model. Fixes #736

### LINKED ISSUE #736: Make RDF and XML attribute definition syntax consistent
Maybe this is just crazy extra work that no-one will use, but I find it pretty confusing that in AF9 there are totally different syntaxes to specify how you index object attributes depending on whether the object metadata is persisted as RDF or XML.  It seems like you should be able to specify how you want your object indexed independently of how it's persisted.

Currently in AF9, you define an object using RDF persistence like this:

```ruby
class Book < ActiveFedora::Base
  property :title, predicate: ::RDF::DC.title, multiple: false do |index|
    index.as :stored_searchable
  end
  property :author, predicate: ::RDF::DC.creator, multiple: false do |index|
    index.as :stored_searchable
  end
end
```

And in AF9, you would define the same attributes persisted to XML as follows:
```ruby
class Codex < ActiveFedora::Base
  contains 'descMetadata', class_name: 'CodexMetadata'

  has_attributes :title, datastream: 'descMetadata', multiple: false
  has_attributes :author, datastream: 'descMetadata', multiple: false
end

class CodexMetadata < ActiveFedora::OmDatastream
  set_terminology do |t|
    t.root(path: "fields")
    t.title
    t.author
  end

  def self.xml_template
    Nokogiri::XML.parse("<fields/>")
  end
end
```

It seems awkward to specify the indexing as part of the persistence for XML only.  Further, it seems like historical cruft that RDF attributes are defined using `property` and XML attributes are defined using \has_metadata`. Would it be possible to do something more like:

```ruby
# A proposed revision of XML persistence
class Codex < ActiveFedora::Base
  contains 'descMetadata', class_name: 'CodexMetadata'

  property :title, delegate_to: 'descMetadata', multiple: false do |index|
    index.as :stored_searchable
  end  

  property :author, delegate_to: 'descMetadata', multiple: false do |index|
    index.as :stored_searchable
  end  
end

class CodexMetadata < ActiveFedora::OmDatastream
  set_terminology do |t|
    t.root(path: "fields")
    t.title
    t.author
  end

  def self.xml_template
    Nokogiri::XML.parse("<fields/>")
  end
end
```

In this example, the XML attachment is used to specify the persistence only and the indexing is handled identically regardless of whether RDF or XML is used to persist metadata.  Rather than providing a `predicate` to the `property` you specify the `delegate_to` to identify which XML attachment the term is persisted to.

jcoyne:
You should now add indexing hints to has_attributes by passing a block
similar to how you do it with rdf properties. e.g.:

```ruby
class Book < ActiveFedora::Base
  has_metadata  'descMetadata', type: MODSDatastream
  has_attributes :title, :license, datastream: 'descMetadata' do |index|
    index.as :stored_searchable, :facetable
  end
end
```

### DIFF
--- a/lib/active_fedora/attributes.rb
+++ b/lib/active_fedora/attributes.rb
@@ -144,7 +144,7 @@ module ActiveFedora
         @delegated_attributes = val
       end
 
-      def has_attributes(*fields)
+      def has_attributes(*fields, &block)
         options = fields.pop
         datastream = options.delete(:datastream).to_s
         raise ArgumentError, "You must provide a datastream to has_attributes" if datastream.blank?
@@ -152,6 +152,7 @@ module ActiveFedora
         fields.each do |f|
           create_attribute_reader(f, datastream, options)
           create_attribute_setter(f, datastream, options)
+          add_attribute_indexing_config(f, &block) if block_given?
         end
       end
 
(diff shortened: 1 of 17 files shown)


mjgiarlo:
:+1: slick!

PR #745, opened 2015-04-01
Prevents an object from being loaded to the incorrect class. 

hectorcorrea:
For example, when loading a batch object it should be loaded into a Batch class (and not a File for example.)

Loading an object to the incorrect class could result in loss of data when saving the object back. For example, loading a File into a Batch and then saving the Batch will blow away all the properties of the File that were not present in the Batch (which would be the majority.)

The current version of the code still allows an object to be loaded into a parent class. For example, a BigFile object could be loaded into a File object as long as BigFile *is a subclass* of File. This could still result in loss of data but I wasn't sure I wanted to be too restrictive so I am leaving this open for now. We might want to prevent this altogether in the future.

### DIFF
--- a/lib/active_fedora/relation/finder_methods.rb
+++ b/lib/active_fedora/relation/finder_methods.rb
@@ -38,7 +38,6 @@ module ActiveFedora
       return to_a.find { |*block_args| yield(*block_args) } if block_given?
       options = args.extract_options!
       options = options.dup
-
       cast = if @klass == ActiveFedora::Base && !options.has_key?(:cast)
         true
       else 
@@ -198,12 +197,20 @@ module ActiveFedora
         ActiveFedora::Base
       else
         # The true class may be a subclass of @klass, so always use from_class_uri
-        Model.from_class_uri(has_model_value(resource)) || ActiveFedora::Base
+        resource_class = Model.from_class_uri(has_model_value(resource)) || ActiveFedora::Base
+        unless equivalent_class?(resource_class)
+          raise ActiveFedora::ActiveFedoraError.new("Model mismatch. Expected #{@klass}. Got: #{resource_class}") 
+        end
+        resource_class
       end
     end
 
     def has_model_value(resource)
       resource.graph.query([nil, ActiveFedora::RDF::Fcrepo::Model.hasModel, nil]).first.object.to_s
+    end
+
+    def equivalent_class?(other_class)
+      other_class <= @klass
     end
 
     def find_with_ids(ids, cast)
--- a/spec/integration/associations_spec.rb
+++ b/spec/integration/associations_spec.rb
@@ -286,6 +286,39 @@ describe ActiveFedora::Base do
       end
     end
 
+    describe "when fetching an existing object" do
+      before do
+        class Dog < ActiveFedora::Base; end
+        class BigDog < Dog; end
+        class Cat < ActiveFedora::Base; end
+        @dog = Dog.create
+        @big_dog = BigDog.create
+      end
+      it "should detect class mismatch" do
+        expect {
+          Cat.find @dog.id
+        }.to raise_error(ActiveFedora::ActiveFedoraError)
+      end
+
+      it "should not accept parent class into a subclass" do
+        expect {
+          BigDog.find @dog.id
+        }.to raise_error(ActiveFedora::ActiveFedoraError)
+      end
+
+      it "should accept a subclass into a parent class" do
+        # We could prevent this altogether since loading a subclass into
+        # a parent class (a BigDog into a Dog) would result in lost of 
+        # data if the object is saved back and the subclass has more 
+        # properties than the parent class. However, it does seem reasonable 
+        # that people might want to use them interchangeably (after all  
+        # a BigDog is a Dog) and therefore we allow for it.
+        expect {
+          Dog.find @big_dog.id
+        }.not_to raise_error    
+      end
+    end
+
     describe "setting belongs_to" do
       before do
         @library = Library.new()


lib/active_fedora/relation/finder_methods.rb
```diff
@@ -206,6 +209,10 @@ def has_model_value(resource)
       resource.graph.query([nil, ActiveFedora::RDF::Fcrepo::Model.hasModel, nil]).first.object.to_s
     end
 
+    def equivalent_class?(other_class)
+      other_class == @klass || other_class < @klass
```
mjgiarlo: How about:

```ruby
def equivalent_or_subclass?(other_class)
  other_class <= @klass
end
```

lib/active_fedora/relation/finder_methods.rb
```diff
@@ -198,7 +197,11 @@ def class_to_load(resource, cast)
         ActiveFedora::Base
       else
         # The true class may be a subclass of @klass, so always use from_class_uri
-        Model.from_class_uri(has_model_value(resource)) || ActiveFedora::Base
+        resource_class = Model.from_class_uri(has_model_value(resource)) || ActiveFedora::Base
```
mjgiarlo: Do you think some folks might like to override the `ActiveFedora::Base` disjunct here?  If not, nothing to see here.  If so:

```ruby
# ...
resource_class = Model.from_class_uri(has_model_value(resource)) || default_class
# ...

def default_class
  ActiveFedora::Base
end
````

File: spec/integration/associations_spec.rb [hunk 3]
```diff
@@ -286,6 +286,35 @@ class Publisher < ActiveFedora::Base
       end
     end
 
+    describe "when fetching an existing object" do
+      before do
+        class Dog < ActiveFedora::Base; end
+        class BigDog < Dog; end
+        class Cat < ActiveFedora::Base; end
+        @dog = Dog.create()
```
mjgiarlo: You can omit empty parens here.

spec/integration/associations_spec.rb
```diff
@@ -286,6 +286,35 @@ class Publisher < ActiveFedora::Base
...
+        @bigDog = BigDog.create()
```
mjgiarlo: :bomb: the empty parens, and prefer snake_case over camelCase for variable names. :snake: > :camel:

spec/integration/associations_spec.rb
```diff
@@ -286,6 +286,35 @@ class Publisher < ActiveFedora::Base
...
+      end
+
+      it "should not accept parent class into a subclass" do
+        expect {
+          BigDog.find(@dog.id)
+        }.to raise_error(ActiveFedora::ActiveFedoraError)
+      end
+
+      it "should accept a subclass into a parent class" do
+        # We could prevent this altogether since loading a subclass into
```
mjgiarlo: In your estimation, should we prevent this or allow it?

spec/integration/associations_spec.rb
```diff
@@ -286,6 +286,35 @@ class Publisher < ActiveFedora::Base
...
+        # a parent class (a bigDog into a Dog) would result in lost of 
+        # data if the subclass had more properties than the parent class
+        # and the object is saved back. 
+        expect(Dog.find(@bigDog.id).id).to eq @bigDog.id
```
mjgiarlo: Do we need to compare identifiers here, or is it sufficient that `Dog.find` doesn't raise here (which brings symmetry with its related tests above)?

awead:
This would prevent loading a class as a different one, yes?
``` ruby
obj = Foo.create!
morphed_obj = Bar.find(obj.id)
```
Some folks actually rely on this behavior. In the past, we discussed making this a validation in the model so you could turn in on or off.

mjgiarlo:
Good point, @awead. I believe the folks at BPL (@scande3 ?) make use of this ability, so it'd be good to hear from him.

awead:
yes, and @dchandekstark as well, I believe. He's been asking about this.

hectorcorrea:
Yes, this PR will prevent the morphed object example that @awead mentioned. 

If this behavior is allowed by design then I can close this PR.

spec/integration/associations_spec.rb
hectorcorrea: I think we should prevent it, but I don't know much about how others use ActiveFedora (particularly given Adam's note that some people might already rely on the current behavior).

dchandekstark:
@barmintor has expressed the strongest dissent on this limitation.  I like most all of @mjgiarlo line comments (the one I could grasp at a glance).  Thanks for cc, @awead - I would like to see this backported also to 7.x and 8.x. :)

mjgiarlo:
I'm neither :+1: or :-1: on changing this behavior, I should say.  We don't currently rely on it for Sufia, so I'd defer to others who either really really need this behavior to stick around or really really object to it.

barmintor:
I'm happy to say that one must use Rubydora directly to change the associated AF model of an object in the future.

mjgiarlo:
@barmintor so you're cool with this PR in AF?

mjgiarlo:
Btw, if this object morphing capability is a thing we want to allow and support, shouldn't we have tests in AF that show this working? (Which these changes ought to have broken, but didn't?)

barmintor:
I'm a solid +0 at this point.

awead:
@mjgiarlo it was never an expressed feature, I just knew that folks like me took advantage of it from time to time. FWIW, it looks like a non-issue in Fedora 4, so I'm :+1:

scande3:
@mjgiarlo and @awead - We never load an object of a type that isn't specified to be. What we do is use class inheritance... so if I have "SimpleObjectBase" and "Photograph", I can find / load the item as either a "SimpleObjectBase" or a "Photograph". I'd never load it as a "Book" as that wouldn't be stored in the "has_model_ssim" in Solr. The only exception is if something changed from a "Photograph" to a "Book" that we have a method to handle that conversion (but I'm unsure of how this would be done in ActiveFedora 9 yet).

From the hydra-tech thread by Stuart Kenny, the current ActiveFedora 9 has no concept of multiple "has_model_ssim" and assumes all objects are of a single class model once again. I plan to work on some way to support this during this weekend but have no issues with an enforcement that an object must have an appropriate "has_model_ssim" assigned to it (assuming there are other ways in ActiveFedora 9 to convert objects from one model to another).

Overall - until I have had more time to look at ActiveFedora 9 and how it currently operates, can't comment on this code implementation other than mentioning our use case above.

mjgiarlo:
Thoughts on this, @jcoyne?  I think all folks here have chimed in with +1 or +0 so far.

lib/active_fedora/relation/finder_methods.rb
hectorcorrea: I didn't change this.

lib/active_fedora/relation/finder_methods.rb
hectorcorrea: Done, nice trick.

spec/integration/associations_spec.rb
hectorcorrea: They are gone now!

jcoyne:
+1

mjgiarlo:
:+1:, :watch: on Travis

PR #747, opened 2015-04-02
Add a mechanism to set rdf_label on the ActiveTriple resource

jcoyne:


mjgiarlo:
:+1:

terrellt:
@jcoyne Could you just extend https://github.com/ActiveTriples/ActiveTriples/blob/master/lib/active_triples/configurable.rb and then populate the resource with that? Although we may need to add something to AT to keep track of what configuration options were set.

jcoyne:
@terrellt I don't think that's possible, because this is on the AF::Base, not on a Resource. I'd be happy if you proved me wrong though.

terrellt:
@jcoyne Using the AT API would look like this: https://github.com/projecthydra/active_fedora/compare/rdf_label...terrellt:rdf_label

Ideally https://github.com/projecthydra/active_fedora/pull/747/files#diff-4827a9cc4420720ae18bf842f60f18f5R94 could be replaced by a method in AT.

The only concern I would have is method name collision.

jcoyne:
@terrellt nice idea, but it's not backwards compatible with 9.0.4

lib/active_fedora/fedora_attributes.rb
```diff
@@ -80,10 +89,14 @@ def resource_class
...
       end
 
-      # Set the rdf type for this class
-      def type(uri)
-        @type = uri
-      end
+      private
+        # @return a Hash of options suitable for passing to ActiveTriples::Base.configure
+        def active_triple_options
+          {}.tap do |h|
```
jeremyf: Why is this getting `tap`ped for hash generation? Its a bit confusing.

Why not `{ type: type, rdf_label: rdf_label }`?

spec/unit/base_spec.rb
```diff
@@ -51,6 +51,50 @@ class FooHistory < ActiveFedora::Base
...
+      after do
+        Object.send(:remove_const, :Person)
+        Object.send(:remove_const, :Agent)
+        Object.send(:remove_const, :Creator)
+      end
+
+      let(:instance) { Creator.new(foaf_name: ['Carolyn'], job: ['Developer']) }
+      subject { instance.rdf_label }
+
+      it { is_expected.to eq ['Developer'] }
```
jeremyf: I'm confused as to why the `instance.rdf_label` would be equal to `['Developer']`? I think I understand why, but I believe there might be another test to help convey the differences.

Is the following correct?

```ruby
agent = Agent.new(foaf_name: ['Carolyn'])
expect(agent.rdf_label).to eq(['Carolyn'])
```

Perhaps two tests to convey how the rdf_label changes as you continue to extend objects.

jcoyne: You example is correct. We have the test on line 69 that illustrates that case.

jeremyf:
:+1: though waiting on @terrellt for either a response or to push the Merge button

PR #748, opened 2015-04-02
The indexing hints should be inheritable

jcoyne:


mjgiarlo:
:ok_hand:

jcoyne:
@jeremyf thanks, I think you found a bug. Fixed and added a test.

PR #749, opened 2015-04-03
Sort versions as dates not as strings

mjgiarlo:


cjcolvar:
:+1:

hectorcorrea:
Looks good to me.

lib/active_fedora/versions_graph.rb
```diff
@@ -2,11 +2,11 @@ module ActiveFedora
...
     def all opts={}
-      if opts[:include_auto_save]
-        fedora_versions
-      else
-        fedora_versions.reject { |v| v.label.match("auto") }
+      versions = fedora_versions
+      unless opts[:include_auto_save]
+        versions.reject! { |version| version.label.match("auto") }
       end
+      versions.sort_by { |version| DateTime.parse(version.created) }
```
hectorcorrea: What happens here is one `version.created` cannot be parsed as a valid date?

mjgiarlo: :bomb:

jcoyne: It seems reasonable to raise an exception if Fedora is giving us a non-date here.

mjgiarlo: Fixed in latest.

hectorcorrea:
:+1: Waiting on Travis....

PR #750, opened 2015-04-03
Clean Connection

terrellt:
This is an LDP Connection that doesn't return server managed/LDP/AF added triples. Useful for returning what you put into LDP.

### DIFF
--- a/lib/active_fedora/clean_connection.rb
+++ b/lib/active_fedora/clean_connection.rb
@@ -1,0 +1,40 @@
+module ActiveFedora
+  class CleanConnection < SimpleDelegator
+    def get(*args)
+      result = __getobj__.get(*args) do |req|
+        prefer_headers = Ldp::PreferHeaders.new(req.headers["Prefer"])
+        prefer_headers.omit = prefer_headers.omit | omit_uris
+        req.headers["Prefer"] = prefer_headers.to_s
+      end
+      CleanResult.new(result)
+    end
+
+    private
+
+    def omit_uris
+      [
+        RDF::Fcrepo4.ServerManaged,
+        RDF::Ldp.PreferContainment,
+        RDF::Ldp.PreferEmptyContainer,
+        RDF::Ldp.PreferMembership
+      ]
+    end
+
+    class CleanResult < SimpleDelegator
+      def graph
+        @graph ||= clean_graph
+      end
+
+      private
+
+      def clean_graph
+        __getobj__.graph.delete(has_model_query)
+        __getobj__.graph
+      end
+
+      def has_model_query
+        [nil, ActiveFedora::RDF::Fcrepo::Model.hasModel, nil]
+      end
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


lib/active_fedora/prefer_headers.rb
```diff
@@ -0,0 +1,59 @@
+module ActiveFedora
+  class PreferHeaders
```
jcoyne: I think we should move this to the LDP Gem. What do you say?

spec/integration/clean_connection_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+
+RSpec.describe ActiveFedora::CleanConnection do
+  subject { ActiveFedora::CleanConnection.new(connection) }
+  let(:connection) { ActiveFedora.fedora.connection }
+  describe "#get" do
+    context "when given an existing resource uri" do
+      let(:uri) { asset.rdf_subject }
+      let(:asset) do
```
mjgiarlo: How about:

```ruby
ActiveFedora::Base.create do |a|
  a.resource << [a.rdf_subject, RDF::DC.title, 'test']
end
```

:question:

terrellt: :+1: Didn't know you could do that.

lib/active_fedora/prefer_headers.rb
terrellt: Seems reasonable, may take a bit.

lib/active_fedora/clean_connection.rb
```diff
@@ -0,0 +1,35 @@
...
+        RDF::Fcrepo4.ServerManaged,
+        RDF::Ldp.PreferContainment,
+        RDF::Ldp.PreferEmptyContainer,
+        RDF::Ldp.PreferMembership
+      ]
+    end
+
+    class CleanResult < SimpleDelegator
+      def graph
+        __getobj__.graph.delete(has_model_query)
```
jeremyf: This is a bit odd. Does the `graph#delete` return a graph object that no longer has the specified triple?

I would expect:

```ruby
def graph
  __getobj__.graph.delete(has_model_query)
  __getobj__.graph
end
```

This change does not rely on the overloaded `graph#delete` method that is acting as both a command and a query method.

terrellt: It is, it does, and sure - I can change that.

terrellt: @jeremyf Look better? Thought it would be good for it not to continuously rerun the delete as well.

One thing that's interesting is graph#delete returning a graph without the statement is similar to how a functional language would do it, no? It -does- mutate the underlying object, but I wonder if the return value was intentional. https://github.com/ruby-rdf/rdf/blob/develop/lib/rdf/mixin/mutable.rb#L137 looks like it is, but immutability certainly doesn't seem to be the goal given the name of the mixin.

jeremyf:
:+1:

PR #755, opened 2015-04-06
Raise an error when the inverse relationship can not be found.

jcoyne:
Previously a StackTooDeep was encountered in this situation


mjgiarlo:
:+1: 

Waiting on Travis.

PR #756, opened 2015-04-07
Derive a foreign_key ending with `_ids` if the inverse is a collection

jcoyne:


jeremyf:
:+1: pending spec completion.

PR #757, opened 2015-04-07
Only set/save the inverse on a HABTM if the inverse is also HABTM

jcoyne:
previously we were getting:

```
     ActiveFedora::UnknownAttributeError:
            Item does not have an attribute `item_id'
```

if the inverse was a has_many.

### DIFF
--- a/lib/active_fedora/associations/has_and_belongs_to_many_association.rb
+++ b/lib/active_fedora/associations/has_and_belongs_to_many_association.rb
@@ -18,13 +18,16 @@ module ActiveFedora
         owner[reflection.foreign_key] ||= []
         owner[reflection.foreign_key] += [record.id]
 
-        if @owner.new_record? and @reflection.options[:inverse_of]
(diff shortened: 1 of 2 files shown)


cjcolvar:
I don't quite understand all that is going on here, but it looks like activerecord doesn't do anything with inverse_of on HABTM relationships (https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/associations.rb#L1641-L1678).  If we're trying to mostly track activerecord's API, do we even need this?

jcoyne:
@cjcolvar we need this because out implementation of HABTM is substantially different from ARs implmentation. In ActiveFedora, we use RDF which can have multiple outgoing assertions vs ActiveRecord using a join_table.  This commit adds no new features, it merely prevents a bug when the inverse is a has_many instead of another HABTM.

dchandekstark:
Wouldn't it be incorrect for the inverse of a HABTM to be a has_many?

jcoyne:
@dchandekstark not in ActiveFedora.  The inverse of a HABTM may be a has_many, which is to say the relationships are only outbound, or another HABTM, which means the relationships are bi-directional.   A HABTM may also be without an inverse relationship.

cjcolvar:
:+1:

PR #758, opened 2015-04-07
Content model inheritance

stkenny:
Most specific model class should be loaded, by using an inheritance structure if present.

### DIFF
--- a/spec/integration/belongs_to_association_spec.rb
+++ b/spec/integration/belongs_to_association_spec.rb
@@ -10,6 +10,9 @@ describe ActiveFedora::Base do
       belongs_to :library, predicate: ActiveFedora::RDF::Fcrepo::RelsExt.hasConstituent
     end
     class SpecialInheritedBook < Book
+      def assert_content_model
+        self.has_model = [self.class.to_s, self.class.superclass.to_s]
+      end
     end
 
   end
@@ -73,8 +76,8 @@ describe ActiveFedora::Base do
     end
 
     it "should cast to the most specific class for the association" do
-      @library2.books[0].class == Book
-      @library2.books[1].class == SpecialInheritedBook
+      expect(@library2.books[0]).to be_instance_of Book
+      expect(@library2.books[1]).to be_instance_of SpecialInheritedBook
     end
 
     after do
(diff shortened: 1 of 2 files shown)


scande3:
+1 to have this merged soon from the BPL. I've tested @stkenny's fix from his branch yesterday and it restored how we use inherited content models. His change also just uses "model matching logic" that already exists in other parts of the codebase that has functioned without causing any issues in the past.

Thanks!

spec/integration/belongs_to_association_spec.rb
```diff
@@ -90,35 +93,61 @@ class SimpleObject < ActiveFedora::Base
         belongs_to :complex_collection, predicate: ActiveFedora::RDF::Fcrepo::RelsExt.isPartOf, class_name: 'ComplexCollection'
       end
 
-      class ComplexObject < SimpleObject
+      #NOTE: As Active Fedora does alphabetical ordering of RDF datastreams automatically, the "Z" is to insure this is stored
```
jcoyne: I'm not sure that this is Active-Fedora doing this. Could it be Fedora returning things in that order? Can you show me where that is happening in the code?

spec/integration/belongs_to_association_spec.rb
```diff
@@ -90,35 +93,61 @@ class SimpleObject < ActiveFedora::Base
...
       Object.send(:remove_const, :SimpleCollection)
       Object.send(:remove_const, :ComplexCollection)
     end
 
     describe "saving between the before and after hooks" do
+
+      context "Verify that an inherited object will save and retrieve with the correct models" do
+        before do
+          @complex_object = ZComplexObject.create
+          @complex_object = ZComplexObject.find(@complex_object.id)
```
jcoyne: Let's prefer using `let` over instance variables. e.g. `let(:complex_object) { ZComplexObject.find(ZComplexObject.create.id) }`

spec/integration/belongs_to_association_spec.rb
```diff
@@ -90,35 +93,61 @@ class SimpleObject < ActiveFedora::Base
...
+
+          @complex_collection = ComplexCollection.create
+          @complex_collection = ComplexCollection.find(@complex_collection.id)
```
jcoyne: Here too

spec/integration/belongs_to_association_spec.rb
```diff
@@ -188,6 +217,56 @@ class ComplexCollection < SimpleCollection
...
+    after :all do
+      Object.send(:remove_const, :SuperclassObject)
+      Object.send(:remove_const, :SubclassObject)
+      Object.send(:remove_const, :SuperclassCollection)
+    end
+
+    describe "Adding subclass objects" do
+      context "Add a subclass_object into a superclass_collection" do
+        before do
+          @superclass_collection = SuperclassCollection.create
```
jcoyne: Please use `let` here. e.g 
```ruby
let(:subclass_object) { SubclassObject.create }
let(:superclass_collection) { SuperclassCollecion.create { |sc| sc.objects = [subclass_object] } }
```

spec/integration/belongs_to_association_spec.rb
```diff
@@ -188,6 +217,56 @@ class ComplexCollection < SimpleCollection
...
+          @superclass_collection.save!
+        end
+        it "should have added the inverse relationship for the correct class" do
+          expect(@subclass_object.superclass_collection).to be_instance_of SuperclassCollection
+        end
+      end
+
+      context "Set the superclass_collection of a subclass object"
+        before do
+          @superclass_collection = SuperclassCollection.create
```
jcoyne: Here too

jcoyne:
Looks pretty good. I added a few suggestions on test style. Then please squash your commits into a single commit.

jcoyne:
@stkenny I'm not seeing your name on the list of people who have a Hydra Contributor License Agreement on file.  Please see https://wiki.duraspace.org/display/hydra/Hydra+Project+Intellectual+Property+Licensing+and+Ownership for more information about filling out an iCLA for yourself and a cCLA for your institution. We are unable to accept your submission without a license agreement on file. (cc @mjgiarlo)

stkenny:
I've sent both the iCLA and cCLA now. Thanks.

spec/integration/belongs_to_association_spec.rb
scande3: Beyond just that is how it is behaving when you persist the data (ie. if you have two content models on an object, it will always find the alphabetically first one regardless of write order), I'm not exactly sure where it is occurring in this version of the codebase.

In the past, RDF::RDFXML::Writer was used when serializing XML that would have this behavior as its default. I can see that the library that includes that is still in the Gemfile (s.add_dependency "rdf-rdfxml", '~> 1.1.0') but do not see its Writer functionality being used from a codebase search. 

So while it is loading these has_model relationships in alphabetical order, I'd need to dig to see where that is occurring now. Regardless: one can remove the note if wanted (+1 to it). Even if the assertion is incorrect, RDF is not guaranteed to be in any order and the implementation detail is a coincidence. This change is still recommended as though as the only reason several tests failed to fail is because "ComplexObject" came before "SimpleObject" in the alphabet and the current behavior was that automatic re-ordering. You can make just only the naming change in the current branch (before merging this) to see that the tests then would have caught the error. 

There are a few additional corrections and tests added now that will fail in the case the RDF write and read are truly random. But a simple change to help ensure some other tests work in the current implementation of ActiveFedora.

jcoyne: Yea, I'm just picking nits about the precision of the statement. How about saying that "RDF assertions seem to be returned in alphabetical order..."

scande3: Sure, sounds good to me. :+1:

mark-dce:
FYI - Trinity College cCLA and iCLA for Stuart Kenny received by Hydra-Legal on 4/7/2015 - expect them to be processed and posted to contributor list shortly.

mjgiarlo:
@stkenny is now clear to contribute to Hydra!  CLAs have been processed.

jcoyne:
@stkenny please change the phrasing of that comment and then we can merge.

stkenny:
Which comment do I need to change?

jcoyne:
@stkenny spec/integration/belongs_to_association_spec.rb:96 See https://github.com/projecthydra/active_fedora/pull/758#discussion_r27880969

PR #772, opened 2015-04-16
Use delegate_to instead of datastream in the options for property

jcoyne:
Trying to reduce/elimnate the useage of "datastream" especially in the
public API.


awead:
:+1: if Travis is ok.

PR #777, opened 2015-04-27
Patches casting behavior

### LINKED ISSUE #746: Backport #745 to version 8.x

dchandekstark:
See https://github.com/projecthydra/active_fedora/wiki/Patching-ActiveFedora-7.x-&-8.x-Casting-Behavior
for detailed information on the problem.

The solution offered in this patch preserves the most-specific-class algorithm
of `ActiveFedora::ContentModel.best_model_for` and raises a new execption,
`ActiveFedora::ModelNotAsserted` in `ActiveFedora::Core#adapt_to_cmodel`.

Note that a repository object can still be loaded into a model class not asserted,
for example, by using `.find` with `cast: false`.

Closes #746 (although not really a backport)

### DIFF
--- a/lib/active_fedora/rdf/identifiable.rb
+++ b/lib/active_fedora/rdf/identifiable.rb
@@ -47,7 +47,7 @@ module ActiveFedora::RDF::Identifiable
     # @param [RDF::URI] uri URI that is being looked up.
     def from_uri(uri,_)
       begin
-        self.find(pid_from_subject(uri))
+        ActiveFedora::Base.find(pid_from_subject(uri))
       rescue ActiveFedora::ObjectNotFoundError
         self.ds_specs[resource_datastream.to_s][:type].resource_class.new(uri)
       end
--- a/spec/unit/om_datastream_spec.rb
+++ b/spec/unit/om_datastream_spec.rb
@@ -407,15 +407,15 @@ describe ActiveFedora::OmDatastream do
       @obj = MyObj.new
       @obj.descMetadata.title = 'Foobar'
       @obj.save
+      @obj.reload
     end
     after do
       @obj.destroy
       Object.send(:remove_const, :MyObj)
     end
-    subject { @obj.reload.descMetadata } 
     it "should not load the descMetadata datastream when calling content_changed?" do
       @obj.inner_object.repository.should_not_receive(:datastream_dissemination).with(hash_including(:dsid=>'descMetadata'))
-      subject.should_not be_content_changed
+      @obj.descMetadata.should_not be_content_changed
     end
   end
 end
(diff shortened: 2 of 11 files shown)


dchandekstark:
Attn @awead @cjcolvar @jcoyne

lib/active_fedora/rdf/identifiable.rb
```diff
@@ -47,7 +47,7 @@ def resource_datastream(ds=nil)
     # @param [RDF::URI] uri URI that is being looked up.
     def from_uri(uri,_)
       begin
-        self.find(pid_from_subject(uri))
+        ActiveFedora::Base.find(pid_from_subject(uri))
```
dchandekstark: I think this is the right adjustment.  Somebody tell me otherwise.

spec/unit/om_datastream_spec.rb
```diff
@@ -407,15 +407,15 @@ class MyObj < ActiveFedora::Base
...
     end
     after do
       @obj.destroy
       Object.send(:remove_const, :MyObj)
     end
-    subject { @obj.reload.descMetadata } 
     it "should not load the descMetadata datastream when calling content_changed?" do
       @obj.inner_object.repository.should_not_receive(:datastream_dissemination).with(hash_including(:dsid=>'descMetadata'))
-      subject.should_not be_content_changed
+      @obj.descMetadata.should_not be_content_changed
```
dchandekstark: Better semantics would have fixed subject as in rdf_datastream_spec below.  Speak up if you care.

lib/active_fedora/rdf/identifiable.rb
jcoyne: I'm not sure I understand why you'd need to make this change.

dchandekstark: @jcoyne Because it blows up now if the uri points to an AF object of a different model (i.e., not <= self.class).  Alternatively, I suppose we could use `cast: false` but this seems better to me.

dchandekstark: @jcoyne Example spec is here: https://github.com/projecthydra/active_fedora/blob/8.0-stable/spec/unit/rdf_resource_datastream_spec.rb#L197-L203.  DummyAsset model is not asserted by DummyResource.

awead:
Taking a quick look, I'm :+1: but differ to the others.

cjcolvar:
:+1:

dchandekstark:
@awead @cjcolvar @jcoyne Merge?

cjcolvar:
@jcoyne Have all of your questions been adequately addressed?  I'm happy to merge if jcoyne +1 or +0's it.

jcoyne:
@cjcolvar fine by me.

PR #780, opened 2015-05-02
Add optional prefix to resource URI

awead:
This is based on some needs I've seen recently. I've seen PCDM examples prefix classes with names in the url, ex.
```
/works/work1
/works/work2
```
`works` is just a basic container. You can specify a containing URI for your classes:
``` ruby
GenericWork < ActiveFedora::Base
  uri_prefix "works"
end
```
Creating a new GenericWork will create a uri similar to:
`/fedora/rest/dev/works/ab/cd/ef/abcdef` depending on your id minting scheme.

### DIFF
--- a/lib/active_fedora.rb
+++ b/lib/active_fedora.rb
@@ -88,6 +88,7 @@ module ActiveFedora #:nodoc:
     autoload :NomDatastream
     autoload :NullRelation
     autoload :OmDatastream
+    autoload :Pathing
     autoload :Persistence
     autoload :ProfileIndexingService
     autoload :Property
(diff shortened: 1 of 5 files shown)


jcoyne:
Awead: Please test the behavior when a resource with a URI not in the root_path is loaded by a class that has root_path set.  Ensure the existing URI is not mutated when saving

jcoyne:
Retracting my earlier comment. I think this looks good. :+1:

awead:
I've refactored this and am calling it `uri_prefix` instead of `root_path`.

jcoyne:
@awead :+1: that's a much better name.

PR #781, opened 2015-05-04
Failing test for nested attributes issue

ronan-mch:
Hi,

Here is a couple of tests that demonstrate the nested models problem described in the mail. When creating this PR I discovered that the test fails regardless of whether they're ActionController::Parameters or not, so I guess that part is unimportant. Note also that when debugging this problem locally, I found that the Work object first began to have extra titles after [the creation of the LDP source](https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/persistence.rb#L151) in the Persistence module. I wasn't able to figure out why though.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem 'byebug' unless ENV['TRAVIS']
 group :test do
   gem 'simplecov', require: false
   gem 'coveralls', require: false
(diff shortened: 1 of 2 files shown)


jcoyne:
Superseded by #783

PR #783, opened 2015-05-04
CollectionAssociation should generate a solr query lazily

jcoyne:
When a CollectionAssociation is loaded before the owing object is saved
the solr query is produced without the id of the owner, so all the
objects of the correct type are returned.
Ref #781

### DIFF
--- a/lib/active_fedora/associations/collection_association.rb
+++ b/lib/active_fedora/associations/collection_association.rb
@@ -3,11 +3,6 @@ module ActiveFedora
     class CollectionAssociation < Association #:nodoc:
       attr_reader :proxy
 
-      def initialize(owner, reflection)
-        super
-        construct_query
-      end
-
       # Implements the reader method, e.g. foo.items for Foo.has_many :items
       # @param opts [Boolean, Hash] if true, force a reload
       # @option opts [Symbol] :response_format can be ':solr' to return a solr result.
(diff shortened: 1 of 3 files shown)


lib/active_fedora/associations/collection_association.rb
```diff
@@ -330,11 +326,12 @@ def reset_target!
...
 
         def construct_query
-
-          #TODO use primary_key instead of id
-          clauses = { find_reflection => @owner.id }
-          clauses[:has_model] = @reflection.class_name.constantize.to_class_uri if @reflection.class_name && @reflection.class_name != 'ActiveFedora::Base'
-          @counter_query = @finder_query = ActiveFedora::SolrQueryBuilder.construct_query_for_rel(clauses)
+          @solr_query ||= begin
+            #TODO use primary_key instead of id
+            clauses = { find_reflection => @owner.id }
```
awead: Do we care about hash rockets here?

jcoyne: How do you suggest writing this?

terrellt: @awead It's the only way - 1.9 syntax only works for symbols.

mjgiarlo: What @terrellt said.

PR #786, opened 2015-05-05
Track type as a changed attribute in MetadataNode

jcoyne:


terrellt:
@jcoyne Was there a bug this fixes?

jcoyne:
@terrellt nothig that I logged, but yes, type wasn't getting set on files.

jcoyne:
Hold off on merging this. It needs one more thing.

PR #788, opened 2015-05-06
Add direct containers

jcoyne:
You can use the containers by calling:
```ruby
class FooHistory < ActiveFedora::Base
   directly_contains :files, predicate:
   ::RDF::URI.new("http://example.com/hasFiles"), class_name: 'Thing'
end
```
if you don't specify a class_name it will default to ActiveFedora::File

### DIFF
--- a/lib/active_fedora/associations/builder/directly_contains.rb
+++ b/lib/active_fedora/associations/builder/directly_contains.rb
@@ -1,0 +1,23 @@
+module ActiveFedora::Associations::Builder
+  class DirectlyContains < CollectionAssociation #:nodoc:
+    self.macro = :directly_contains
+    self.valid_options += [:has_member_relation, :is_member_of_relation]
+    self.valid_options -= [:predicate]
+
+    def build
+      reflection = super
+      configure_dependency
+      reflection
+    end
+
+    def validate_options
+      super
+      if !options[:has_member_relation] && !options[:is_member_of_relation]
+        raise ArgumentError, "You must specify a predicate for #{name}"
+      elsif !options[:has_member_relation].kind_of?(RDF::URI) && !options[:is_member_of_relation].kind_of?(RDF::URI)
+        raise ArgumentError, "Predicate must be a kind of RDF::URI"
+      end
+    end
+  end
+end
+
--- a/lib/active_fedora/associations/directly_contains_association.rb
+++ b/lib/active_fedora/associations/directly_contains_association.rb
@@ -1,0 +1,60 @@
+module ActiveFedora
+  module Associations
+    class DirectlyContainsAssociation < CollectionAssociation #:nodoc:
+
+      def insert_record(record, force = true, validate = true)
+        container.save!
+        if force
+          record.save!
+        else
+          record.save(validate: validate)
+        end
+      end
+
+      def reader
+        @records ||= ContainerProxy.new(self)
+      end
+
+      def find_target
+        query_node = if container_predicate = options[:has_member_relation]
+           owner
+        else
+          container_predicate = ::RDF::Vocab::LDP.contains
+          container
+        end
+
+        uris = query_node.resource.query(predicate: container_predicate).map { |r| r.object.to_s }
+
+        uris.map { |object_uri| klass.find(klass.uri_to_id(object_uri)) }
+      end
+
+      def container
+        @container ||= begin
+          DirectContainer.find_or_initialize(ActiveFedora::Base.uri_to_id(uri)).tap do |container|
+            container.parent = @owner
+            container.has_member_relation = Array(options[:has_member_relation])
+            container.is_member_of_relation = Array(options[:is_member_of_relation])
+          end
+        end
+      end
+
+      protected
+
+        def count_records
+          load_target.size
+        end
+
+        def initialize_attributes(record) #:nodoc:
+          record.uri = ActiveFedora::Base.id_to_uri(container.mint_id)
+          set_inverse_instance(record)
+        end
+
+      private
+
+        def uri
+          raise "Can't get uri. Owner isn't saved" if @owner.new_record?
+          "#{@owner.uri}/#{@reflection.name}"
+        end
+    end
+  end
+end
--- a/lib/active_fedora/containers/container.rb
+++ b/lib/active_fedora/containers/container.rb
@@ -1,0 +1,30 @@
+module ActiveFedora
+  # This is the base class for ldp containers, it is not an ldp:BasicContainer
+  class Container < ActiveFedora::Base
+
+    property :membership_resource, predicate: ::RDF::Vocab::LDP.membershipResource
+    property :has_member_relation, predicate: ::RDF::Vocab::LDP.hasMemberRelation
+    property :is_member_of_relation, predicate: ::RDF::Vocab::LDP.isMemberOfRelation
+
+    def parent
+      @parent || raise("Parent hasn't been set on #{self.class}")
+    end
+
+    def parent=(parent)
+      @parent = parent
+      self.membership_resource = [::RDF::URI(parent.uri)]
+    end
+
+    def mint_id
+      "#{id}/#{SecureRandom.uuid}"
+    end
+
+    def self.find_or_initialize(id)
+      find(id)
+    rescue ActiveFedora::ObjectNotFoundError
+      new(id)
+    end
+  end
+end
+
+
--- a/spec/integration/direct_container_spec.rb
+++ b/spec/integration/direct_container_spec.rb
@@ -1,0 +1,151 @@
+require 'spec_helper'
+
+describe "Direct containers" do
+  describe "#directly_contains" do
+    context "when the class is ActiveFedora::File" do
+      before do
+        class FooHistory < ActiveFedora::Base
+           directly_contains :files, has_member_relation: ::RDF::URI.new("http://example.com/hasFiles")
+        end
+      end
+      after do
+        Object.send(:remove_const, :FooHistory)
+      end
+
+      let(:file) { o.files.build }
+      let(:reloaded) { FooHistory.find(o.id) }
+
+      context "with no files" do
+        let(:o) { FooHistory.new }
+        subject { o.files }
+
+        it { is_expected.to be_empty }
+        it { is_expected.to eq [] }
+      end
+
+      context "when the object exists" do
+        let(:o) { FooHistory.create }
+
+        before do
+          file.content = "HMMM"
+          o.save
+        end
+
+        describe "#first" do
+          subject { reloaded.files.first }
+          it "has the content" do
+            expect(subject.content).to eq 'HMMM'
+          end
+        end
+
+        describe "#to_a" do
+          subject { reloaded.files }
+          it "has the content" do
+            expect(subject.to_a).to eq [file]
+          end
+        end
+
+        describe "#append" do
+          let(:file2) { o.files.build }
+          it "has two files" do
+            expect(o.files).to eq [file, file2]
+          end
+
+          context "and then saved/reloaded" do
+            before do
+              file2.content = "Derp"
+              o.save!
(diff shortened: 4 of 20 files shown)


awead:
Since @hectorcorrea and @terrellt are already looking at this, I'm giving this a tentative :+1: as long as it can create things that look like what's been outlined in https://github.com/awead/sufia-pcdm. I'm happy to give this a try, if you want.

spec/integration/direct_container_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+        end
+
+        describe "#first" do
+          subject { reloaded.files.first }
+          it "has the content" do
+            expect(subject.content).to eq 'HMMM'
+          end
+        end
+
+        describe "#to_a" do
```
terrellt: Should we have a test for what happens when there are no files?

lib/active_fedora/containers/container.rb
```diff
@@ -0,0 +1,28 @@
+module ActiveFedora
+  class Container < ActiveFedora::Base
```
terrellt: Maybe a comment that makes it clear that this isn't an object representation of a BasicContainer

lib/active_fedora/containers/container.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    property :membership_resource, predicate: ::RDF::Vocab::LDP.membershipResource
```
terrellt: Validations for these necessary?

lib/active_fedora/associations/directly_contains_association.rb
```diff
@@ -0,0 +1,54 @@
...
+        return true
+      end
+
+      def reader
+        @records ||= ContainerProxy.new(self)
+      end
+
+      def find_target
+        uris = owner.resource.query(predicate: options[:predicate]).map { |r| r.object.to_s }
+        if klass <= ActiveFedora::File # a subclass of file
```
terrellt: I thought the operator for this was `<`

lib/active_fedora/associations/directly_contains_association.rb
```diff
@@ -0,0 +1,54 @@
...
+          uris.map { |uri| klass.new(uri) }
```
terrellt: ActiveFedora::File doesn't respond to #find?

jcoyne: Nope

lib/active_fedora/associations/directly_contains_association.rb
jcoyne: I want same as or a subclass of, so: `<=`

lib/active_fedora/containers/container.rb
jcoyne: It's already validated. If you remove this line the tests fail.

lib/active_fedora/associations/directly_contains_association.rb
terrellt: Can we make it?

lib/active_fedora/containers/container.rb
terrellt: Because of LDP?

lib/active_fedora/containers/container.rb
```diff
@@ -0,0 +1,28 @@
...
+    def parent=(parent)
+      @parent = parent
+      self.membership_resource = [::RDF::URI(parent.uri)]
+    end
+
+    def mint_id
+      "#{id}/#{SecureRandom.uuid}"
+    end
+
+    def self.find_or_initialize(id)
```
terrellt: Should this be on AF::Base?

lib/active_fedora/containers/container.rb
terrellt: Ah, no, because it's only using Container through that relation. I could do Container.new.save! and it'd run fine

jcoyne: because: `NoMethodError:  undefined method `membership_resource='`

lib/active_fedora/containers/container.rb
jcoyne: I don't see a reason to move it there unless there's anyone else who would want to use this.

terrellt: I guess I'm not sure what's special about Containers that would make us want to find_and_initialize here but not there.

lib/active_fedora/containers/container.rb
terrellt: I meant a presence validation.

lib/active_fedora/associations/directly_contains_association.rb
jcoyne: That seems like extra work, because we'd do a uri -> id -> uri for no benefit.

lib/active_fedora/containers/container.rb
jcoyne: I don't understand how that would happen. We're always building it correctly, so why have a check that costs cycles and brings no benefit?

lib/active_fedora/containers/container.rb
jcoyne: It's a very specific method and is hardly ever useful. We've received complaints about the API being overly broad, so I don't want to expand it unless you can show me a case where you would use it.

spec/integration/direct_container_spec.rb
jcoyne: Added. See line 18

hectorcorrea:
@jcoyne This implementation only address the scenario in which the direct container is created with a ` hasMemberRelation`. 

In other words, it does not let you create direct containers with `isMemberOfRelation` instead of `hasMemberRelation`, right? 

I am asking because we might need support for `isMemberRelation` in the implementation of PCDM (see here: https://wiki.duraspace.org/display/FEDORA4x/LDP-PCDM-F4+In+Action#LDP-PCDM-F4InAction-Ordering-CreateDirectContainer)

jcoyne:
@hectorcorrea I'm not seeing where isMemberRelation is used in PCDM (in the link you sent me). Can you help me find that?

lib/active_fedora/associations/directly_contains_association.rb
terrellt: I think there's a lot of benefit in just sending messages and not having to do a class check to get our functionality. Happy to be overridden.

hectorcorrea:
@jcoyne Look for 

```
Like the "pages/" DirectContainer in an earlier example, the "orderProxies/" includes the ldp:membershipResource property ("raven/"). However, it is important to point out that the "orderProxies/" DirectContainer *does not* have the ldp:hasMemberRelation property defined, but instead uses ldp:isMemberOfRelation of "ore:proxyIn".
By using ldp:isMemberOfRelation, the auto-created triple resulting from the addition of a new child resource within "orderProxies/" will take the form

<new-resource> <ore:proxyIn> <http://localhost:8080/fcrepo/rest/objects/raven/>
```

hectorcorrea:
@jcoyne This PR is +1 for me, though. We could handle the scenario of supporting `isMemberOfRelation` as a separate PR.

hectorcorrea:
@jcoyne I didn't realize the Duraspace Wiki page does not open all the child pages by default. Search for "Ordering In Action" instead, open the "Ordering Creation Details..." link, and then you'll see the text that I mentioned under "Ordering - Create DirectContainer"

lib/active_fedora/containers/container.rb
hectorcorrea: +1 for keeping the API narrow.

no-reply: I'm also :+1: on keeping the API small, but that seems like a motivation to make this method private, or to consider a builder pattern, not to tuck it away in the subclass.  My concern would be that the outcome here is that we end up with an API that is both big *and* inconsistent.

jcoyne:
@hectorcorrea the latest version lets you set `isMemberOfRelation`

lib/active_fedora/containers/container.rb
jcoyne: @no-reply I don't know what that means. Care to make a pull request or show me what you're talking about?

jcoyne: @no-reply Specifically in relation to the original comment which said to move this method to AF::Base. Seems like keeping it just on containers keeps the AF::Base api smaller.

lib/active_fedora/associations/builder/directly_contains.rb
```diff
@@ -0,0 +1,23 @@
...
+    def build
+      reflection = super
+      configure_dependency
+      reflection
+    end
+
+    def validate_options
+      super
+      if !options[:has_member_relation] && !options[:is_member_of_relation]
+        raise ArgumentError, "You must specify a predicate for #{name}"
```
hectorcorrea: {name} ?

lib/active_fedora/associations/directly_contains_association.rb
```diff
@@ -0,0 +1,64 @@
...
+  module Associations
+    class DirectlyContainsAssociation < CollectionAssociation #:nodoc:
+
+      def insert_record(record, force = true, validate = true)
+        container.save!
+        if record.new_record?
+          if force
+            record.save!
+          else
+            return false unless record.save(validate: validate)
```
hectorcorrea: shouldn't `return record.save(validate: validate)` be clearer?

lib/active_fedora/associations/builder/directly_contains.rb
jcoyne: ?

lib/active_fedora/associations/directly_contains_association.rb
jcoyne: Fixed.

spec/integration/direct_container_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+      before do
+        class FooHistory < ActiveFedora::Base
+           directly_contains :files, has_member_relation: ::RDF::URI.new("http://example.com/hasFiles")
+        end
+      end
+      after do
+        Object.send(:remove_const, :FooHistory)
+      end
+
+      let(:file) { o.files.build }
```
hectorcorrea: I found it somewhat non-intuitive to use `build` rather than something line `add` or `create` but I guess since `build` is the ActiveRecord standard we should go with it. (sincerely, the the department of pointless comments)

hectorcorrea:
@jcoyne The `isMemberOfRelation` implementation looks great. I documented[1] the questions that we had yesterday over IRC about how the containment is represented when using `isMemberOfRelation` vs `hasMemberRelation` in case somebody wonders about it later. 

[1] https://gist.github.com/hectorcorrea/b7c80ce00914040c5ed7

lib/active_fedora/associations/builder/directly_contains.rb
hectorcorrea: where is name defined?

jcoyne: In the superclass

jcoyne: Oh, yes, it is in the superclass: https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/associations/builder/association.rb#L9

lib/active_fedora/containers/container.rb
terrellt: @jcoyne I mean, this keeps the AF::Base API smaller, but just trades that concern with an inconsistent access API, no?

I'm not willing to hold this up just for this, but it was a concern

terrellt:
I'm :+1: at this point. Any lingering issues?

lib/active_fedora/containers/container.rb
no-reply: Yeah, this is not a blocker. It's a philosophical/design difference that we should probably resolve elsewhere.

PR #797, opened 2015-05-12
Validation callback

ronan-mch:
As I understand it, the `before_validation` callback should be called before the `save` event and the `valid?` check. As far as I can see, this doesn't happen with `ActiveFedora::Base`. My apologies if I'm misunderstanding something.

### DIFF
--- a/spec/unit/callback_spec.rb
+++ b/spec/unit/callback_spec.rb
@@ -70,7 +70,7 @@ describe ActiveFedora::Base do
 
   end
 
-  describe 'validations bug' do
+  describe 'before_validation callback' do
     module SpecModel
       class Book < ActiveFedora::Base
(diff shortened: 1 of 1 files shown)


spec/unit/callback_spec.rb
```diff
@@ -69,4 +69,28 @@ def do_stuff
     @cb.destroy
 
   end
+
+  describe 'before_validation callback' do
+    module SpecModel
```
jcoyne: put this module declaration in a `before` block. Add an `after` block that removes the module. That should allow the test suite to pass.

atz: Alternatively, use an anonymous module: `Module.new do; ...; end`

PR #799, opened 2015-05-13
When Translator is changed on a subclass it should not affect the parent

jcoyne:


terrellt:
I'm a little scared to allow translators to vary by class. Won't that break auto-casting?

jcoyne:
If we don't, I'm seeing errors like this: 
https://travis-ci.org/projecthydra/sufia/jobs/62410621

jcoyne:
I think #803 is maybe a better way to fix this given @terrellt's concerns.

PR #803, opened 2015-05-13
Translation procs should not overwrite one another

jcoyne:
mattr_accessor with a block sets the class variable when it's called.
If the value is customized and then Identifiable is included on another
class mattr_accessor resets the custom value to the default. This commit
prevents that overwriting.

See:
https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/module/attribute_accessors.rb#L72

This supersedes #799


terrellt:
:+1: 

@jcoyne asked for my blessing, so :sparkles: :sparkles:

PR #810, opened 2015-06-16
Add versioning to File

jpstroop:
`File` was including `Versionable` but not calling `has_many_versions`. Assuming that `File`s should be versionable, this PR fixes.

### DIFF
--- a/lib/active_fedora/file.rb
+++ b/lib/active_fedora/file.rb
@@ -10,6 +10,9 @@ module ActiveFedora
     extend ActiveTriples::Properties
     generate_method 'content'
 
+    include ActiveFedora::Versionable
+    has_many_versions
+
     extend ActiveModel::Callbacks
     include Identifiable
     include Scoping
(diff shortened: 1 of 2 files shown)


jcoyne:
I don't think we want to automatically version files. The way it was before gives you the option to make them versionable, but does not insist they are.

jpstroop:
@jcoyne is there a way I'm missing to set whether a File is versionable or not on an instance?

jcoyne:
On your concrete class that extends File, call `has_many_versions`.  See https://github.com/projecthydra/sufia/blob/eb714cb91f4ecd49dec67b0e9996dc3d4e918f1a/sufia-models/lib/sufia/models/file_content/versions.rb#L7
and https://github.com/projecthydra/sufia/blob/eb714cb91f4ecd49dec67b0e9996dc3d4e918f1a/sufia-models/app/models/datastreams/file_content_datastream.rb#L3

Issue #794, opened 2015-05-08
Can't delete Direct Container Files


terrellt:
```ruby
o.files = [file, file2]
o.save
o.files = [file] # => Undefined method #persisted? for #<ActiveFedora::File....>
```

jcoyne:
@terrellt fixed?

terrellt:
I haven't tried. I don't think so

PR #816, opened 2015-06-22
(Needs Review) implements directly_contains_one association

flyingzumwalt:
Note: part of this functionality is blocked by #794

### DIFF
--- a/lib/active_fedora/associations.rb
+++ b/lib/active_fedora/associations.rb
@@ -26,6 +26,7 @@ module ActiveFedora
     autoload :HasAndBelongsToManyAssociation, 'active_fedora/associations/has_and_belongs_to_many_association'
     autoload :BasicContainsAssociation,       'active_fedora/associations/basic_contains_association'
     autoload :DirectlyContainsAssociation,    'active_fedora/associations/directly_contains_association'
+    autoload :DirectlyContainsOneAssociation, 'active_fedora/associations/directly_contains_one_association'
     autoload :IndirectlyContainsAssociation,  'active_fedora/associations/indirectly_contains_association'
     autoload :ContainsAssociation,            'active_fedora/associations/contains_association'
     autoload :DeleteProxy,            'active_fedora/associations/delete_proxy'
@@ -43,6 +44,7 @@ module ActiveFedora
       autoload :HasAndBelongsToMany, 'active_fedora/associations/builder/has_and_belongs_to_many'
       autoload :Contains,            'active_fedora/associations/builder/contains'
       autoload :DirectlyContains,    'active_fedora/associations/builder/directly_contains'
+      autoload :DirectlyContainsOne, 'active_fedora/associations/builder/directly_contains_one'
       autoload :IndirectlyContains,  'active_fedora/associations/builder/indirectly_contains'
 
       autoload :Property,         'active_fedora/associations/builder/property'
@@ -106,6 +108,10 @@ module ActiveFedora
       #
       def directly_contains(name, options={})
         Builder::DirectlyContains.build(self, name, { class_name: 'ActiveFedora::File' }.merge(options))
+      end
+
+      def directly_contains_one(name, options={})
+        Builder::DirectlyContainsOne.build(self, name, { class_name: 'ActiveFedora::File' }.merge(options))
       end
 
       # This method is used to declare an ldp:IndirectContainer on a resource
--- a/lib/active_fedora/associations/builder/directly_contains_one.rb
+++ b/lib/active_fedora/associations/builder/directly_contains_one.rb
@@ -1,0 +1,44 @@
+module ActiveFedora::Associations::Builder
+  class DirectlyContainsOne < SingularAssociation #:nodoc:
+    self.macro = :directly_contains_one
+    self.valid_options += [:has_member_relation, :is_member_of_relation, :type, :through]
+    self.valid_options -= [:predicate]
+
+    def validate_options
+      raise ArgumentError, "you must specify a :through option on #{name}.  #{name} will use the container from that directly_contains association." if !options[:through]
+      if options[:through]
+        inherit_options_from_association(options[:through])
+      end
+      super
+
+      if options[:class_name] == "ActiveFedora::File"
+        raise ArgumentError, "You cannot set :class_name of #{name} to ActiveFedora::File because directly_contains_one needs to assert and read RDF.type assertions, which is not supported by ActiveFedora::File.  To make Files support RDF.type assertions, define a subclass of ActiveFedora::File and make it `include ActiveFedora::WithMetadata`. Otherwise, all subclasses of ActiveFedora::Base support RDF.type assertions."
+      elsif !options[:has_member_relation] && !options[:is_member_of_relation]
+        raise ArgumentError, "You must specify a :has_member_relation or :is_member_of_relation predicate for #{name}"
+      elsif !options[:has_member_relation].kind_of?(RDF::URI) && !options[:is_member_of_relation].kind_of?(RDF::URI)
+        raise ArgumentError, "Predicate must be a kind of RDF::URI"
+      end
+
+      if !options[:type].kind_of?(RDF::URI)
+        raise ArgumentError, "You must specify a Type and it must be a kind of RDF::URI"
+      end
+    end
+
+    private
+
+    # Inherits :has_member_relation from the association corresponding to association_name
+    # @param [Symbol] association_name of the association to inherit from
+    def inherit_options_from_association(association_name)
+      associated_through_reflection = lookup_reflection(association_name)
+      raise ArgumentError, "You specified `:through => #{@reflection.options[:through]}` on the #{name} associaiton but #{model} does not actually have a #{@reflection.options[:through]}` association" if associated_through_reflection.nil? || !associated_through_reflection.name
+      raise ArgumentError, "You must specify a directly_contains association as the :through option on #{name}.  You provided a #{associated_through_reflection.macro}" unless associated_through_reflection.macro == :directly_contains
+      options[:has_member_relation] = associated_through_reflection.options[:has_member_relation] unless options[:has_member_relation]
+      options[:class_name] = associated_through_reflection.options[:class_name] unless (options[:class_name] && options[:class_name] != "ActiveFedora::File")
+    end
+
+    def lookup_reflection(association_name)
+      model.reflect_on_association(association_name)
+    end
+
+  end
+end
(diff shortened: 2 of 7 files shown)


flyingzumwalt:
@jcoyne @terrellt I'm not in love with this code, but I need this feature in order to proceed with projecthydra-labs/hydra-works#131

jcoyne:
Looks like this depends on #794? So should we wait until that is fixed?

flyingzumwalt:
directly_contains_one doesn't rely on #794 any more than direclty_contains and indirectly_contains, which are already in master.  I'd like to proceed with adding directly_contains_one so I can continue work on curation_concerns & sufia-core.

lib/active_fedora/associations/builder/directly_contains_one.rb
```diff
@@ -0,0 +1,39 @@
...
+    self.valid_options -= [:predicate]
+
+    def validate_options
+      if options[:through]
+        inherit_options_from_association(options[:through])
+      end
+      super
+
+      if !options[:has_member_relation] && !options[:is_member_of_relation]
+        raise ArgumentError, "You must specify a predicate for #{name}"
```
jcoyne: This error message is a bit misleading. Can is say, specify has_member_relation or is_member_of_relation?

lib/active_fedora/associations/directly_contains_one_association.rb
```diff
@@ -0,0 +1,118 @@
...
+        # filtered_objects = container_association_proxy.to_a.select { |o| o.metadata_node.type.include?(options[:type]) }
+        query_node = if container_predicate = options[:has_member_relation]
+                       owner
+                     else
+                       container_predicate = ::RDF::Vocab::LDP.contains
+                       container_association.container  # Use the :through association's container
+                     end
+
+        contained_uris = query_node.resource.query(predicate: container_predicate).map { |r| r.object.to_s }
+        contained_objects = contained_uris.map { |object_uri| klass.find(klass.uri_to_id(object_uri)) }
```
jcoyne: A more efficient way to do this is:
```ruby
return contained_uris.find do |object_uri|
  contained_object = klass.find(klass.uri_to_id(object_uri))
  contained_object.metadata_node.type.include?(options[:type])
end
```
Then it only needs to load objects until it finds one that matches, rather than load all the objects.

lib/active_fedora/associations/directly_contains_one_association.rb
```diff
@@ -0,0 +1,118 @@
...
+        add_to_container(record)
+        add_type_to_record(record, options[:type])
+        record
+      end
+
+      # Adds record to the DirectContainer identified by the container_association
+      # Relies on container_association.initialize_attributes to appropriately set things like record.uri
+      def add_to_container(record)
+        container_association.add_to_target(record)  # adds record to corresponding Container
+        container_association.send(:initialize_attributes,record) # Uses the :through association initialize the record with things like the correct URI for a direclty contained object
```
jcoyne: Add space after comma.

lib/active_fedora/associations/directly_contains_one_association.rb
```diff
@@ -0,0 +1,118 @@
...
+      end
+
+      # Replaces association +target+ with +record+
+      def replace(record, save = true)
```
jcoyne: Second argument is never used. can you change it to `*` if it is required?

lib/active_fedora/associations/directly_contains_one_association.rb
```diff
@@ -0,0 +1,118 @@
...
+      def add_type_to_record(record, type_uri)
+        if record.respond_to?(:metadata_node)
+          metadata_node = record.metadata_node
+        else
+          metadata_node = record
+        end
+        types = metadata_node.get_values(:type)
+        return file if types.include?(type_uri)
+        types << type_uri
+        record.metadata_node.set_value(:type, types)
```
jcoyne: What if `record` isn't a File, then it won't have metadata_node.

lib/active_fedora/associations/directly_contains_one_association.rb
```diff
@@ -0,0 +1,118 @@
...
+        query_node = if container_predicate = options[:has_member_relation]
+                       owner
+                     else
+                       container_predicate = ::RDF::Vocab::LDP.contains
+                       container_association.container  # Use the :through association's container
+                     end
+
+        contained_uris = query_node.resource.query(predicate: container_predicate).map { |r| r.object.to_s }
+        contained_objects = contained_uris.map { |object_uri| klass.find(klass.uri_to_id(object_uri)) }
+        filtered_objects = contained_objects.select {|o| o.metadata_node.type.include?(options[:type]) }
```
jcoyne: What if contained_objects aren't `ActiveFedora::File`s, then they won't respond to `metadata_node`.

lib/active_fedora/associations.rb
```diff
@@ -26,6 +26,7 @@ module Associations
     autoload :HasAndBelongsToManyAssociation, 'active_fedora/associations/has_and_belongs_to_many_association'
     autoload :BasicContainsAssociation,       'active_fedora/associations/basic_contains_association'
     autoload :DirectlyContainsAssociation,    'active_fedora/associations/directly_contains_association'
+    autoload :DirectlyContainsOneAssociation,    'active_fedora/associations/directly_contains_one_association'
```
jcoyne: please fix indenting so it lines up.

lib/active_fedora/associations.rb
```diff
@@ -43,6 +44,7 @@ module Builder
       autoload :HasAndBelongsToMany, 'active_fedora/associations/builder/has_and_belongs_to_many'
       autoload :Contains,            'active_fedora/associations/builder/contains'
       autoload :DirectlyContains,    'active_fedora/associations/builder/directly_contains'
+      autoload :DirectlyContainsOne,    'active_fedora/associations/builder/directly_contains_one'
```
jcoyne: fix indenting so it lines up.

lib/active_fedora/associations/directly_contains_one_association.rb
```diff
@@ -0,0 +1,118 @@
...
+        else
+          remove_existing_target
+        end
+
+        self.target = record
+      end
+
+      # Sets record as the target of the association
+      # Ensures that the RDF.type is set on the target
+      def target=(record)
```
jcoyne: Why implement this? Shouldn't we just move `add_type_to_record` into `replace`?

lib/active_fedora/associations/directly_contains_one_association.rb
```diff
@@ -0,0 +1,118 @@
...
+        add_type_to_record(record, options[:type])
+        @target = record
+      end
+
+      def remove_existing_target
```
jcoyne: This method should be private

lib/active_fedora/associations/directly_contains_one_association.rb
```diff
@@ -0,0 +1,118 @@
...
+        @target ||= find_target
+        if @target
+          container_association_proxy.delete @target
+          @updated = true
+        end
+      end
+
+      # Overrides build_record to ensure that record is initialized with attributes from the corresponding container
+      def build_record(attributes)
```
jcoyne: This method should be private.

lib/active_fedora/associations/directly_contains_one_association.rb
```diff
@@ -0,0 +1,118 @@
...
+        # TODO make initalize take a block and get rid of the tap
+        reflection.build_association(attributes).tap do |record|
+          container_association.initialize_attributes(record) # initializes record with attributes from the corresponding container
+          initialize_attributes(record)
+        end
+      end
+
+      # Returns the Reflection corresponding to the direct container association that's being filtered
+      def container_reflection
```
jcoyne: Make this private

lib/active_fedora/associations/directly_contains_one_association.rb
```diff
@@ -0,0 +1,118 @@
...
+        @container_reflection ||= @owner.class.reflect_on_association(@reflection.options[:through])
+      end
+
+      # Returns the DirectContainerAssociation corresponding to the direct container that's being filtered
+      def container_association
```
jcoyne: Make this private

lib/active_fedora/associations/directly_contains_one_association.rb
```diff
@@ -0,0 +1,118 @@
...
+        container_association_proxy.proxy_association
+      end
+
+      # Returns the ContainerAssociationProxy corresponding to the direct container that's being filtered
+      def container_association_proxy
```
jcoyne: Make this private

lib/active_fedora/associations/directly_contains_one_association.rb
```diff
@@ -0,0 +1,118 @@
...
+
+      # # Build/Create record and ensure that it's added to the corresponding Container
+      def new_record(method, attributes={})
+        # This works, using the container's builder, but then you're stuck with the container's class_name
+        # record = @owner.send(container_reflection.name).send(method, attributes) do |record|
+        #   add_type_to_record(record, options[:type])
+        # end
+
+        # Build the record using _this_ association's build method (so setting things like class_name on this association will have an effect)
+        record = @reflection.send("#{method}_association", attributes)
```
jcoyne: Do we need to override this method? It seem that the super will call replace which should add_to_container. See comment below (line 60) regarding moving `add_type_to_record` to `replace` too.

spec/integration/directly_contains_one_association_spec.rb
```diff
@@ -0,0 +1,108 @@
+require 'spec_helper'
+
+describe ActiveFedora::Base do
+  before do
```
jcoyne: I'd like to see a test here that uses `directly_contains` and a class that extends `ActiveFedora::Base` (instead of just `ActiveFedora::File`)

lib/active_fedora/associations/directly_contains_one_association.rb
flyingzumwalt: SingularAssociation#create and SingularAssociation#build both call new_record without ever explicitly calling replace, so I overrode `new_record` but now I see that SingularAssociation#new_record calls `set_new_record`, which calls `replace`.  What a mess. Updating DirectlyContainsOneAssociation#replace

lib/active_fedora/associations/directly_contains_one_association.rb
terrellt: Do `ActiveFedora::File`s respond to `#resource`? If so, that could be a common interface.

spec/integration/directly_contains_one_association_spec.rb
flyingzumwalt: directly_contains breaks when you use ActiveFedora::Base instead of ActiveFedora::File.  I'll submit an issue and a PR with tests to show the problem.

flyingzumwalt: See #820

PR #817, opened 2015-06-23
Refactoring DelegatedAttribute

awead:
Day One of ActiveFedora Conclave. This refactors DelegatedAttribute into three different subclasses for better code organization.

### DIFF
--- a/lib/active_fedora/attributes.rb
+++ b/lib/active_fedora/attributes.rb
@@ -177,7 +177,7 @@ module ActiveFedora
         elsif properties.key?(:delegate_to)
           define_delegated_accessor([name], properties.delete(:delegate_to), properties.reverse_merge(multiple: true), &block)
         else
-          raise "You must provide `:datastream' or `:predicate' options to property"
+          raise "You must provide `:delegate_to' or `:predicate' options to property"
         end
       end
 
@@ -186,7 +186,7 @@ module ActiveFedora
       def define_active_triple_accessor(name, properties, &block)
         warn_duplicate_predicates name, properties
         properties = { multiple: true }.merge(properties)
-        find_or_create_defined_attribute(name, nil, properties)
+        find_or_create_defined_attribute(name, ActiveTripleAttribute, properties)
         raise ArgumentError, "#{name} is a keyword and not an acceptable property name." if protected_property_name? name
         reflection = ActiveFedora::Attributes::PropertyBuilder.build(self, name, properties, &block)
         ActiveTriples::Reflection.add_reflection self, name, reflection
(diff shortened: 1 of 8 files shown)


awead:
@hectorcorrea actually did most of this work, I just rebased it with some minor refactoring and adding documentation.

lib/active_fedora/delegated_attribute.rb
```diff
@@ -2,105 +2,12 @@ module ActiveFedora
   # Represents the mapping between a model attribute and a field in a datastream
```
cam156: We need to update this comment.  This no longer represents a field in a stream, but a field in any object

cjcolvar:
I'm confused as to the special treatment of ActiveTripleAttribute versus the rest of the attributes subclassing from DelegatedAttribute.  Is it just convenience to use :predicate instead of :delegate_to or is that a requirement of how AT works?  It is a bit confusing from a naming/code reading perspective for ActiveTripleAttributes to be added to delegated_attributes yet not be DelegatedAttributes.

jcoyne:
@cjcolvar they are delegated attributes, they are just delegated to an ActiveTriples resource rather than a Datastream.

cjcolvar:
How about renaming DelegatedAttribute (and associated methods) to something like DelegatedFileAttribute to clarify things?

jcoyne:
@cjcolvar  DelegatedAttribute is the superclass of all the attributes.  DelegatedStreamAttribute is the superclass of OM + RDFDatastream attributes.

lib/active_fedora/attributes.rb
```diff
@@ -215,31 +218,46 @@ def warn_duplicate_predicates new_name, new_properties
...
+      # @param [Symbol] field the field to find or create
+      # @param [Class] klass the class to use to delegate the attribute (e.g. 
+      #                ActiveTripleAttribute, OmAttribute, or RdfDatastreamAttribute)
+      # @param [Hash] args 
+      # @option args [String] :delegate_target the path to the delegate
+      # @option args [Class] :klass the class to create
+      # @option args [true,false] :multiple (false) true for multi-value fields
+      # @option args [Array<Symbol>] :at path to a deep node 
+      # @return [DelegatedAttribute] the found or created attribute
+      def find_or_create_defined_attribute(field, klass, args)
```
cjcolvar: It appears that this method is only being used for create and not find anymore.  Should the find aspect be removed?

cjcolvar:
:clap: Looks good.  There are a few more "datastream" references that should probably be changed to "file" that could be addressed in this PR.

lib/active_fedora/attributes.rb
jcoyne: It doesn't create one if one is found. Same as it did before.

PR #818, opened 2015-06-24
Return relation for .limit, fixes #352

### LINKED ISSUE #352: Associations do not allow chaining relations
class Bar < ActiveFedora::Base
  has_many :foos, property: :has_member
end

end

b = Bar.create
b.foos = [Foo.create, Foo.create]
b.foos.limit(1)
NoMethodError: undefined method `limit' for "[#<Foo pid:"changeme:8", >, #<Foo pid:"changeme:9", >]":Array

This applies to all relations: #where, #limit, #order, etc.  It would be nice for this to work so #accessible_by could be chained to associations:
b.foos.accessible_by(current_ability)

awead:
Adds the .scope method to CollectionProxy so that you can chain
relations such as .limit to a CollectionProxy.

### DIFF
--- a/lib/active_fedora/associations/collection_proxy.rb
+++ b/lib/active_fedora/associations/collection_proxy.rb
@@ -842,6 +842,12 @@ module ActiveFedora
         @association
       end
 
+      # @return [Relation] object for the records in this association
+      def scope
+        @association.scope
+      end
+      alias spawn scope
+
       def to_ary
         load_target.dup
       end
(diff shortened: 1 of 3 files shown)


mjgiarlo:
:+1:

awead:
@cjcolvar this should fix your use case. You can merge it if it looks good.

cjcolvar:
:+1: All of the AF9 work must have set the stage for this since that snippet wasn't enough before.  A backport of this seems a bit unlikely then.

PR #820, opened 2015-06-24
BUG: test shows that directly_contains does not support elements that are ActiveFedora::Base

flyingzumwalt:
Fails when the association tries to override its element's .uri method.  ActiveFedora::File lets you set .uri=, but ActiveFedora::Base doesn't.  I'm not sure if there are further failures beyond that.

You can run the relevant tests with 

```
rspec spec/integration/directly_contains_one_association_spec.rb:57
```


flyingzumwalt:
@jcoyne as you requested...

flyingzumwalt:
Updated the branch to also include test coverage within direct_container_spec.rb

PR #825, opened 2015-06-24
Removing unneeded line

cam156:
The line is already being done in the refresh method https://github.com/psu-stewardship/active_fedora/blob/remove_dead/lib/active_fedora/persistence.rb#L166


mjgiarlo:
:+1:

PR #826, opened 2015-06-24
Removing dead code in QueryMethods

awead:
Removing these methods because they aren't being used anymore... at least as far as we can tell.


mjgiarlo:
:+1:

awead:
nevermind. nothing to see here... move along...

PR #827, opened 2015-06-24
Use foreign_key in case user has specified one

awead:


mjgiarlo:
:+1: if Travis is :green_apple:

PR #828, opened 2015-06-24
Make autosave tests more specific

awead:
Autosave should not be included when it's not needed and should be tested to ensure it saves changes to existing objects


mjgiarlo:
:+1: if travis is :relaxed:

PR #829, opened 2015-06-25
Reorganizing ActiveFedora:File code

awead:
This splits out some of the code in ActiveFedora::File into relevant modules for better organization and to
make the File class overall lighter and more easy to read.

### DIFF
--- a/lib/active_fedora/file.rb
+++ b/lib/active_fedora/file.rb
@@ -1,19 +1,27 @@
 require 'deprecation'
 
 module ActiveFedora
-
-  #This class represents a Fedora datastream
   class File
-    include AttributeMethods # allows 'content' to be tracked
-    include ActiveModel::Dirty
-    extend Deprecation
-    extend ActiveTriples::Properties
-    generate_method 'content'
-
     extend ActiveModel::Callbacks
+    extend ActiveSupport::Autoload
+    extend ActiveTriples::Properties
+    extend Deprecation
+    extend Querying
+
+    autoload :Streaming
+    autoload :Attributes
+
+    include ActiveFedora::File::Attributes
+    include ActiveFedora::File::Streaming
+    include ActiveFedora::Persistence
+    include ActiveFedora::Versionable
+    include ActiveModel::Dirty
+    include AttributeMethods # allows 'content' to be tracked
     include Identifiable
     include Scoping
-    extend Querying
+
+    generate_method 'content'
+
     define_model_callbacks :save, :create, :destroy
     define_model_callbacks :initialize, only: :after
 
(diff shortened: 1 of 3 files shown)


lib/active_fedora/file/fedora_attributes.rb
```diff
@@ -0,0 +1,39 @@
+module ActiveFedora::File::FedoraAttributes
```
jcoyne: I like this module. I'm not so keen on the name. I don't think that Fedora has anything do do with whether or not a File has a mime_type, name, digest, etc. Can we just have `AF::File::Attributes`?

lib/active_fedora/file/streaming.rb
```diff
@@ -0,0 +1,43 @@
+module ActiveFedora::File::Streaming
+  # @param range [String] the Range HTTP header
+  # @return [Stream] an object that responds to each
+  def stream(range = nil)
+    uri = URI.parse(self.uri)
+    FileBody.new(uri, headers(range, authorization_key))
+  end
+
+  # @return [String] current authorization token from Ldp::Client
+  def authorization_key
```
jcoyne: I think this should be private.

lib/active_fedora/file/fedora_attributes.rb
```diff
@@ -0,0 +1,39 @@
...
+    response.map(&:object)
+  end
+
+  protected
+
+    def default_mime_type
+      'text/plain'
+    end
+
+    def fetch_mime_type
```
jcoyne: I think these two methods can be private.  Also, can we change the name of one of them so they both either use `_from_headers` or both don't?

lib/active_fedora/file.rb
```diff
@@ -111,25 +119,10 @@ def remote_content
...
-    def original_name
-      @original_name ||= fetch_original_name_from_headers
-    end
-
-    def digest
-      response = metadata.ldp_source.graph.query(predicate: ActiveFedora::RDF::Fcrepo4.digest)
-      response.map(&:object)
-    end
-
     def persisted_size
```
jcoyne: I would think this belongs with the original_name, digest, mime_type, etc.

lib/active_fedora/file.rb
```diff
@@ -210,26 +199,12 @@ def readonly?
...
-        URI.decode(m[:filename])
-      end
-
-      def fetch_mime_type
-        ldp_source.head.headers['Content-Type']
-      end
-
     private
 
       def links
```
jcoyne: `links` should probably go to File::FedoraAttributes too.

awead:
@jcoyne it hath been amended and rebased to thy biddings.

Issue #809, opened 2015-06-15
Should AF objects as property values be indexed?


terrellt:
```ruby
class MyObject < ActiveFedora::Base
  property :another_object, :predicate => RDF::DC.related, :class_name => "AnotherObject" do |index|
    index.as :stored_searchable
  end
end

class AnotherObject < ActiveFedora::Base
end

a = MyObject.new
a.related = [AnotherObject.create]
a.to_solr[:related_tesim] # => nil
# I'd want it to be the other object's URI.
```

This seems to have been done purposely at https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/rdf/indexing_service.rb#L89-L92

Could we make that not happen?

PR #830, opened 2015-06-25
RDF::IndexingService indexes objects & properties

awead:
Solrizing AF::Base objects and properties share common behaviors. This refactors both methods for indexing objects and properties into a common method: `solrize_rdf_assertions` which creates other classes that can be overridden if users wish to alter the ways in which their solr fields are created.

This also addresses #809 where objects can be indexed using their uris as the value.

### DIFF
--- a/lib/active_fedora/rdf/field_map.rb
+++ b/lib/active_fedora/rdf/field_map.rb
@@ -1,0 +1,90 @@
+module ActiveFedora::RDF
+  # Transient class that maps solr field names, without their suffixes, to the values and behaviors that
+  # are used to transforming them for insertion into the solr document.
+  # It partially extends Ruby's Hash class, similar to the way ActiveFedora::Indexing::Map does,
+  # but only with selected methods as outlined in def_delegators.
+  class FieldMap
+    extend Forwardable
+    
+    def_delegators :@hash, :[], :[]=, :each, :keys
+    
+    def initialize(hash = {}, &block)
+      @hash = hash
+      yield self if block_given?
+    end
+
+    # Inserts each solr field map configuration into the FieldMap class
+    # @param [Symbol] name the name of the property on the object that we're indexing
+    # @param [Object] index_field_config an instance of ActiveFedora::Indexing::Map::IndexObject
+    # @param [Object] object the instance of ActiveFedora::Base which is being indexed into Solr
+    def insert(name, index_field_config, object)
+      self[index_field_config.key.to_s] ||= FieldMapEntry.new
+      PolymorphicBuilder.new(self[index_field_config.key.to_s], index_field_config, object, name).build
+    end
+
+    # Supports assigning the delegate class that calls .build to insert the fields into the solr document.
+    # @attr [Object] entry instance of ActiveFedora::RDF::FieldMapEntry which will contain the values of the solr field
+    # @attr [Object] index_field_config an instance of ActiveFedora::Indexing::Map::IndexObject
+    # @attr [Object] object the instance of ActiveFedora::Base which is being indexed into Solr
+    # @attr [Symbol] name the name of the property on the object that we're indexing
+    class PolymorphicBuilder
+
+      attr_accessor :entry, :index_field_config, :object, :name
+
+      def initialize(entry, index_field_config, object, name)
+        @entry              = entry
+        @index_field_config = index_field_config
+        @object             = object
+        @name               = name
+        self
+      end
+
+      def build
+        delegate_class.new(entry, index_field_config, object, name).build
+      end
+
+      private
+
+        def delegate_class
+          kind_of_af_base? ? ResourceBuilder : PropertyBuilder
+        end
+
+        def kind_of_af_base?
+          config = properties[name.to_s]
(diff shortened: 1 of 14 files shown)


awead:
@terrellt yo :dog2:, I hear you like indexing.

terrellt:
@awead Did you put indexing in my indexing?

lib/active_fedora/rdf/indexing_service.rb
```diff
@@ -64,35 +64,107 @@ def resource
...
+          index_config.each { |name, index_field_config| field_map.insert(name, index_field_config, object) }
         end
-        field_map
       end
 
-      def kind_of_af_base?(name)
-        config = properties[name.to_s]
-        config && config[:class_name] && config[:class_name] < ActiveFedora::Base
+      # Override this method to use your own FieldMap class for custom indexing of objects and properties
+      def field_map_class
```
terrellt: I'm welcome to be overridden, but this is a textbook factory. `field_map_factory`?

lib/active_fedora/rdf/indexing_service.rb
```diff
@@ -64,35 +64,107 @@ def resource
...
+        def_delegators :@hash, :[], :[]=, :each, :keys
+        
+        def initialize(hash = {}, &block)
+          @hash = hash
+          yield self if block_given?
+        end
+
+        def insert(name, index_field_config, object)
+          self[index_field_config.key.to_s] ||= FieldMapEntry.new
+          builder = kind_of_af_base?(name, object) ? resource_builder_class : property_builder_class
```
terrellt: Can you extract this logic to a polymorphic builder factory?

terrellt:
There's a lot of nested value objects in the rdf indexing service which are potentially useful in other contexts. Can they be extracted?

lib/active_fedora/rdf/indexing_service.rb
terrellt: To clarify, I'm imagining something like this...
```ruby
class PolymorphicBuilder
  def build(entry, index_field_config, object, name)
    delegate_class(name, object).build(entry, index_field_config, object, name)
  end

  private

  def delegate_class(name, object)
    if kind_of_af_base?(name, object)
      ResourceBuilder
    else
      PropertyBuilder
    end
  end

  def kind_of_af_base(name, object)
    config = properties(object)[name.to_s]
    config && config[:class_name] && config[:class_name] < ActiveFedora::Base
  end

  def properties(object)
    object.class.properties
  end
end
```

When you write it out this way you start to see some patterns emerging too - object/name are passed around a lot, it seems like those may be something you'd want in a constructor.

lib/active_fedora/rdf/indexing_service.rb
jcoyne: Doesn't this just give you the problem of "Which factory class should I instantiate?" `FactoryFactory`?

terrellt: Not sure I understand

jcoyne: Neither do I. Can you give an example of what you want it changed to?

awead:
@terrellt I'm not sure what you mean about the factory stuff, but I took your PolymorphicBuilder and refactored with that and split out the classes to their own files.

terrellt:
@awead It was just a method name comment, but we can leave it as is for consistency with the rest of the code base. I had one last note on the polymorphic builder, but after that discussion and potential changes I'm :+1:

terrellt:
Oh, I lied. Documentation comments incoming.

lib/active_fedora/rdf/field_map.rb
```diff
@@ -0,0 +1,80 @@
...
+    extend Forwardable
+    
+    def_delegators :@hash, :[], :[]=, :each, :keys
+    
+    def initialize(hash = {}, &block)
+      @hash = hash
+      yield self if block_given?
+    end
+
+    def insert(name, index_field_config, object)
```
terrellt: Documentation please.

lib/active_fedora/rdf/field_map.rb
```diff
@@ -0,0 +1,80 @@
...
+
+        def properties
+          object.class.properties
+        end
+
+    end
+
+    # Abstract class for building FieldMap entries
+    class Builder
+      def self.build(entry, index_field_config, object, name)
```
terrellt: Documentation please.

lib/active_fedora/rdf/field_map_entry.rb
```diff
@@ -0,0 +1,24 @@
...
+  class FieldMapEntry
+
+    attr_accessor :type, :behaviors, :values
+
+    def initialize
+      @behaviors = []
+      @values = []
+    end
+
+    def merge!(type, behaviors, values)
```
terrellt: Documentation please.

awead:
@terrellt add the documentation. I also smoothed out the PolymorphicBuild/Builder interface so that they now are the same. You can also chain the methods now, like: `PolymorphicBuilder.new(a,b,c,d).build`

PR #832, opened 2015-06-25
Implements delete for direct containers.

### LINKED ISSUE #794: Can't delete Direct Container Files

hectorcorrea:
Fixes #794

### DIFF
--- a/lib/active_fedora/associations/collection_association.rb
+++ b/lib/active_fedora/associations/collection_association.rb
@@ -365,8 +365,12 @@ module ActiveFedora
           existing_records = records.select { |r| r.persisted? }
 
           records.each { |record| callback(:before_remove, record) }
+
+          # Delete the record from Fedora.
           delete_records(existing_records, method) if existing_records.any?
+
           records.each do |record|
+            # Remove the record from the array/collection.
             target.delete(record)
           end
 
--- a/lib/active_fedora/file.rb
+++ b/lib/active_fedora/file.rb
@@ -221,6 +221,10 @@ module ActiveFedora
         refresh
       end
 
+      def build_ldp_resource(id)
+        build_ldp_resource_via_uri self.class.id_to_uri(id)
+      end
+
       def build_ldp_resource_via_uri(uri=nil, content='')
         Ldp::Resource::BinarySource.new(ldp_connection, uri, content, ActiveFedora.fedora.host + ActiveFedora.fedora.base_path)
       end
--- a/spec/integration/direct_container_spec.rb
+++ b/spec/integration/direct_container_spec.rb
@@ -147,6 +147,58 @@ describe "Direct containers" do
         end
       end
     end
+
+    context "deleting members" do
+      before do
+        class FooHistory < ActiveFedora::Base
+           directly_contains :files, is_member_of_relation: ::RDF::URI.new("http://example.com/isWithin")
+        end
+      end
+      after do
+        Object.send(:remove_const, :FooHistory)
+      end
+
+      describe "#delete the contained object directly" do
+        let(:history) { FooHistory.create }
+        let(:file1) { history.files.build }
+        let(:file2) { history.files.build }
+
+        before do
+          file1.content = "hello"
+          file2.content = "hola"
+          history.save
+          history.reload
+        end
+
+        it "deletes the contained resource directly" do
+          expect(history.files).to eq [file1, file2]
+          file1.delete
+          history.reload
+          expect(history.files).to eq [file2]
+        end
+      end
+
+      describe "#delete via the collection proxy" do
+        let(:history) { FooHistory.create }
+        let(:file1) { history.files.build }
+        let(:file2) { history.files.build }
+
+        before do
+          file1.content = "hello"
+          file2.content = "hola"
+          history.save
+          history.reload
+        end
+
+        it "deletes the contained resource via the collection proxy" do
+          expect(history.reload.files).to eq [file1, file2]
+          history.files.delete(file1)
+          expect(history.reload.files).to eq [file2]
+        end
+      end
+
+    end
+
   end
 
   describe "#include?" do
--- a/lib/active_fedora/associations.rb
+++ b/lib/active_fedora/associations.rb
@@ -74,6 +74,14 @@ module ActiveFedora
       association
     end
 
+    def delete(*)
+      reflections.each_pair do |name, reflection|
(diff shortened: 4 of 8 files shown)


flyingzumwalt:
Could you also add a test for this behavior in direct_container_spec.rb?

hectorcorrea:
@flyingzumwalt I added a couple of tests for direct_container and that uncovered a few bugs. 

I still need to (1) make sure the new tests that I added are correct and excising the right functionality, and (2) figure out why 4 of the integration tests are now failing.

spec/integration/direct_container_spec.rb
```diff
@@ -147,6 +147,58 @@ class FooHistory < ActiveFedora::Base
...
+        before do
+          file1.content = "hello"
+          file2.content = "hola"
+          history.save
+          history.reload
+        end
+
+        it "deletes the contained resource directly" do
+          expect(history.files).to eq [file1, file2]
+          file1.delete
```
hectorcorrea: @jcoyne could you confirm is this how we want to test the delete in direct containers? (1/2)

spec/integration/direct_container_spec.rb
```diff
@@ -147,6 +147,58 @@ class FooHistory < ActiveFedora::Base
...
+        before do
+          file1.content = "hello"
+          file2.content = "hola"
+          history.save
+          history.reload
+        end
+
+        it "deletes the contained resource via the collection proxy" do
+          expect(history.reload.files).to eq [file1, file2]
+          history.files.delete(file1)
```
hectorcorrea: @jcoyne this is the other way I am testing delete. (2/2)

lib/active_fedora/associations/collection_association.rb
```diff
@@ -367,7 +367,11 @@ def delete_or_destroy(records, method)
 
           records.each { |record| callback(:before_remove, record) }
           delete_records(existing_records, method) if existing_records.any?
-          records.each do |record|
+          # Existing_records are deleted in the line above, therefore
+          # I am excluding them from the loop below. This seems to be
+          # causing problems with 4 integration tests. Still need to
+          # figure out why.
+          (records-existing_records).each do |record|
```
hectorcorrea: @jcoyne I updated this loop to only process records that have not already been deleted. Yet, this is now making 4 integration tests fail. 

Does this change make sense? I updated this loop because one of the new tests was throwing an LDP::Gone when it was looping to see if one of the records to delete was new. The LDP::Gone was because the record was deleted in the call to `delete_records` just above the loop. If this change make sense I can see why the integration tests are failing but wanted to check before I go down the rabbit hole.

spec/integration/direct_container_spec.rb
jcoyne: Not quite. This is just deleting the file, but not through the association.

spec/integration/direct_container_spec.rb
jcoyne: Yes, I believe this is what we want tested.

lib/active_fedora/associations/collection_association.rb
jcoyne: I don't even know why this loop is necessary if we are deleting the records on line 369.

hectorcorrea: I figured out the rational for this code. The call to `delete_records()` issues the delete to Fedora/Solr. The call to `target.delete()` removes the element from the array/collection.

lib/active_fedora/file.rb
```diff
@@ -63,7 +63,15 @@ def ldp_connection
     # If this file has a parent with ldp#contains, we know it is not new.
     # By tracking exists we prevent an unnecessary HEAD request.
     def new_record?
-      !@exists && ldp_source.new?
+      !@exists && ldp_new_record?
```
hectorcorrea: @jcoyne this is what I needed to do to get the delete for direct_container to work. The call to `target.delete(record)` in `collection_association.rb` internally calls `new_record?` **after** the actual record has been deleted. I believe when Ruby processes target.delete(record) it's using the `==` operator defined in file.rb which in turns calls new_record.

jcoyne: I wonder if we should be tracking `@destroyed` like core does? (https://github.com/projecthydra/active_fedora/blob/7ed4ca1cb4443b6b39dccea4dba18442809cb028/lib/active_fedora/persistence.rb#L49) Can you show me where `target.delete(record)` is calling `==`?

hectorcorrea: Tracking `@destroyed` might be another way to do it.

There is no explicit place in the code where *we* call `==` but when I step into the `target.delete(record)` line it takes me directly to the `==` so I *suspect* Ruby is internally calling `==` when deleting an element in an array.

jcoyne: So, where is `target.delete()`? in `CollectionAssociation`?

hectorcorrea: Yes, https://github.com/projecthydra/active_fedora/pull/832/files#diff-c3a5b16cb7738636967c4263617d88ffR375

jcoyne: I'm not sure we need `==` to use `new_record?`. I'm working up a new PR to remove that. Please stand by.

jcoyne: @hectorcorrea try rebasing against https://github.com/projecthydra/active_fedora/pull/840. Then would you be able to revert your change to new_record?

hectorcorrea: @jcoyne that did it!

hectorcorrea:
@jcoyne @flyingzumwalt I think this PR is good to go (despite coverall's meaningless warning)

Eventually I would like to add another test in `direct_container_spec.rb` to test `directly_contains` without the use of files but I won't have time to get to this for a while. So I would suggest merging this PR as-is.

lib/active_fedora/file.rb
```diff
@@ -221,6 +221,10 @@ def update_record(options = {})
         refresh
       end
 
+      def build_ldp_resource(id)
```
jcoyne: Can you remove this? I don't see it used anywhere.

jcoyne: Actually, now I recall that we did have some reason to add it. Never mind.

PR #837, opened 2015-06-27
Avoid unnecessary solr query

jcoyne:
If only one property has been defined for a predicate,
there's no reason to run a query to filter the results by class.
This avoids a call to solr that is unnecessary in most situations.

### DIFF
--- a/spec/integration/associations/rdf_spec.rb
+++ b/spec/integration/associations/rdf_spec.rb
@@ -1,0 +1,49 @@
+require 'spec_helper'
+
+describe "rdf associations" do
+  context "when there is one relationship for the predicate" do
+    before do
+      class Foo < ActiveFedora::Base
+      end
+      class Library < ActiveFedora::Base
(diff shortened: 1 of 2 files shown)


spec/integration/associations/rdf_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      class Bar < ActiveFedora::Base
+      end
+      class Library < ActiveFedora::Base
+        has_and_belongs_to_many :foos, predicate: ::RDF::URI('http://example.com')
+        has_and_belongs_to_many :bars, predicate: ::RDF::URI('http://example.com')
+      end
+    end
+    after do
+      Object.send(:remove_const, :Foo)
+      Object.send(:remove_const, :Library)
```
awead: remove const. Bar here.

jcoyne: I'm now removing Bar on line 38, in the context where it is defined.

PR #840, opened 2015-06-29
Refactor File#== to avoid an unnecessary API call

jcoyne:
The new_record? call used to cause an API call to fedora, which we are
now able to avoid.


grosscol:
No spec for the == operator, but the code is straight forward and checks out on the console.

PR #843, opened 2015-06-30
make CollectionAssociation#find_target private

jcoyne:


cam156:
@jcoyne I wonder is setting find_target to private is going to have ripple effects in other projects.
This may be a breaking change (as with others from last week).

jcoyne:
@cam156 if someone was using find_target, they likely aren't going to get the results they expect as evidenced in the email chain from this morning.  I'll bet you a :beer: that no trouble will be reported as a result of this bug.

PR #844, opened 2015-06-30
Refactor CollectionAssociation#reset

jcoyne:


cam156:
@jcoyne where are the tests?  What does the @stale_state = nil fix?

jcoyne:
@cam156 no tests added because there was no new functionality added. The changed code is covered (see coveralls)

jcoyne:
https://coveralls.io/builds/2940422/source?filename=lib%2Factive_fedora%2Fassociations%2Fcollection_association.rb#L49

jcoyne:
https://coveralls.io/builds/2940422/source?filename=lib%2Factive_fedora%2Fassociations%2Fcollection_association.rb#L49

grosscol:
Looks good.  Functionally `Array.new` and `= []` are equivalent.  Uses a call to super to reset instance variables: loaded, target, stale_state, and inversed.

lib/active_fedora/associations/association.rb
```diff
@@ -24,6 +24,7 @@ def initialize(owner, reflection)
       def reset
         @loaded = false
         @target = nil
+        @stale_state = nil
```
cam156: @jcoyne isn't this new?

PR #846, opened 2015-07-02
Remove has_many_versions

jcoyne:
This no longer does anything meaninful, so it has been removed to
prevent confusion.


mjgiarlo:
:+1:

Issue #849, opened 2015-07-09
Prefer quick check of id before full object check


elrayle:
Checking id and class first is more efficient than a full object check, so reverse the order evaluation to check full object equivalence only if id and class check fail.

jcoyne:
Can you prove that? I think object.equal? Is just comparing the address of the object which is about the fastest thing you can do.

elrayle:
I mistakenly was under the impression that the object equivalence test was comparing values.  It is not.  It is checking that the two objects being tested are indeed the exact same object living at the exact same address in memory.

PR #850, opened 2015-07-09
fix 849 - check id/class before object equivalence

elrayle:
Fixes https://github.com/projecthydra/active_fedora/issues/849


elrayle:
Fixes issue https://github.com/projecthydra/active_fedora/issues/849

elrayle:
The assumption that id/class check is more efficient was incorrect.  Closing the pull request.  See issue https://github.com/projecthydra/active_fedora/issues/849 for discussion.

Issue #857, opened 2015-07-14
BUG: directly_contains doesn't let you create relationships in memory and doesn't work with anything but Files 


flyingzumwalt:
I'll submit a PR to show the problem

PR #858, opened 2015-07-14
DON'T MERGE - SHOWS BUG: directly_contains doesn't let you create relationships in memory and doesn't work with anything but Files

flyingzumwalt:
ref #857

3 tests to show what happens when you
1. try to directly contain an ActiveFedora::Base instead of an ActiveFedora::File
2. try to add directly_contained objects in memory (add to a parent object that hasn't been saved yet)

### DIFF
--- a/spec/integration/direct_container_spec.rb
+++ b/spec/integration/direct_container_spec.rb
@@ -2,6 +2,37 @@ require 'spec_helper'
 
 describe "Direct containers" do
   describe "#directly_contains" do
+
+    context "when the class is an ActiveFedora::Base"  do
+      before do
+        class Forest < ActiveFedora::Base
+          directly_contains :trees, has_member_relation: ::RDF::URI.new("http://example.com/hasTrees"), class_name: "ActiveFedora::Base"
+        end
+      end
+      after do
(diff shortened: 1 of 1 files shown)


flyingzumwalt:
@elrayle @jcoyne Clearer examples of the problems with directly_contains that are forcing downstream tests to always create/save instead of testing in memory.

terrellt:
@jcoyne I'm looking at this ticket now and the major issue seems to be that you have to assign an ID at creation. Maybe AF::Base objects should have flexible IDs until they're persisted?

jcoyne:
@terrellt yes, the current solution was an expedient way to get it done, but certainly not the best way.  It probably just needs to get set up with an autosave.

mjgiarlo:
@terrellt Does this PR still need attention?

terrellt:
@mjgiarlo Yup, still broken.

grosscol:
Just ran into this one again as I'm trying to refactor the specs to avoid calling .save when possible.  Manually assigning an id when creating an object doesn't alleviate the situation.

jrgriffiniii:
I am closing this due to the age of this pull request.

PR #860, opened 2015-07-15
Improve handling of imperfect profile_json when loading instances from Solr.

### LINKED ISSUE #856: Load_instance_from_solr has issues if model has properties that aren't saved in profile_json
This issue comes up if you add new properties to your model and then try to load existing objects from Solr that don't have the new properties set in profile_json. If the new property is singular (i.e. has ```multiple: false```) then trying to read it will raise an error. If it's a multi-value property then reading it returns nil instead of an empty list. This only happens when you use load_instance_from_solr. If you load from Fedora then it behaves as expected, singular properties return nil and multi-value properties return an empty list.

I understand that arguably this might not really be a bug since you can reasonably be expected to perform some kind of a migration when changing the model. However, it would certainly be handy during development if new properties could be just simply added. And in any case, the behaviour should probably be consistent between loading from Fedora or Solr.

Below is a test case to demonstrate this. Can be placed in loadable_from_json_spec.rb.

```ruby
  context "unset fields in profile" do
    before do
      class UnsetTest < ActiveFedora::Base
      end
      @unset_test_id = UnsetTest.create.id
      Object.send(:remove_const, :UnsetTest)
      class UnsetTest < ActiveFedora::Base
        property :singular_field, multiple: false, predicate: ::RDF::URI('http://example.org/singular')
        property :multiple_field, multiple: true, predicate: ::RDF::URI('http://example.org/multiple')
      end
    end
    after do
      Object.send(:remove_const, :UnsetTest)
    end

    context "when loading from Solr" do
      subject { UnsetTest.load_instance_from_solr(@unset_test_id) }

      it "singular field should be nil" do
        expect(subject.singular_field).to be_nil
      end

      it "multiple field should be []" do
        expect(subject.multiple_field).to eql( [] )
      end
    end

    context "when loading from Fedora" do
      subject { UnsetTest.find(@unset_test_id) }

      it "singular field should be nil" do
        expect(subject.singular_field).to be_nil
      end

      it "multiple field should be []" do
        expect(subject.multiple_field).to eql( [] )
      end
    end

  end
```

And I also made a simple fix that'll pass the test but I'm not entirely confident that this is the right way to fix this or if I'm overlooking something else. I don't know the inner workings of Active Fedora that well. But if these look ok, I can certainly just make this a pull request.

```ruby
------------------- lib/active_fedora/loadable_from_json.rb -------------------
index d3e47b4..a38edbf 100644
@@ -117,7 +117,19 @@ module ActiveFedora
         attached_files[key] = SolrBackedMetadataFile.new
       end
       @resource = SolrBackedResource.new(self.class)
-      self.attributes = attrs.slice(*self.class.attribute_names)
+      new_attributes = {}
+      self.class.attribute_names.each do |a|
+        if !attrs.key? a
+          if self.class.reflect_on_property(a).multiple?
+            new_attributes[a] = []
+          else
+            new_attributes[a] = nil
+          end
+        else
+          new_attributes[a] = attrs[a]
+        end
+      end
+      self.attributes = new_attributes
       # TODO Should we clear the change tracking, or make this object Read-only?
 
       run_callbacks :find
```

ojlyytinen:
Fixes #856

I reorganised the code a bit compared to what I posted in the issue discussion. spec/integration/solr_instance_loader_spec.rb already had similar tests so I moved my tests there instead.

### DIFF
--- a/lib/active_fedora/loadable_from_json.rb
+++ b/lib/active_fedora/loadable_from_json.rb
@@ -117,7 +117,7 @@ module ActiveFedora
         attached_files[key] = SolrBackedMetadataFile.new
       end
       @resource = SolrBackedResource.new(self.class)
-      self.attributes = attrs.slice(*self.class.attribute_names)
+      self.attributes = adapt_attributes(attrs)
       # TODO Should we clear the change tracking, or make this object Read-only?
 
       run_callbacks :find
@@ -125,6 +125,39 @@ module ActiveFedora
       freeze
       self
     end
+
+    private
+
+      # Adapt attributes read from Solr to possible minor changes in data model
+      # since the attributes were saved.
+      # @param attrs [Hash] attributes read from Solr
+      # @return [Hash] the adapted attributes
+      def adapt_attributes(attrs)
+        self.class.attribute_names.each_with_object({}) do |attribute_name, new_attributes|
+          new_attributes[attribute_name] = adapt_attribute_value(attrs, attribute_name)
+        end
+      end
+
+      # Adapts a single attribute value from the given attributes hash to match
+      # minor changes in the data model.
+      # @param attrs [Hash] attributes read from Solr
+      # @param attribute_name [String] the name of the attribute to adapt
+      # @return [Object] the adapted value
+      def adapt_attribute_value(attrs, attribute_name)
+        reflection = self.class.reflect_on_property(attribute_name)
+        if !reflection
+          return attrs[attribute_name] # if this isn't a property, copy value verbatim
+        else
+          multiple = reflection.multiple?
+          if !attrs.key?(attribute_name)
+            # value is missing in attrs, add [] or nil
+            return multiple ? [] : nil
+          else
+            # convert an existing scalar to an array if needed
+            return multiple ? Array(attrs[attribute_name]) : attrs[attribute_name]
+          end
+        end
+      end
 
   end
 end
--- a/spec/integration/solr_instance_loader_spec.rb
+++ b/spec/integration/solr_instance_loader_spec.rb
@@ -87,12 +87,37 @@ describe ActiveFedora::SolrInstanceLoader do
     end
   end
 
-  context "when the model has extra values in its json" do
-    let(:profile) { { "foo"=>["baz"], "bar"=>"quix", "title"=>"My Title", "extra_value"=>"Bonus values!"}.to_json }
+  context "when the model has imperfect json" do
     let(:doc) { { 'id' => 'test-123', 'has_model_ssim'=>['Foo'], 'object_profile_ssm' => profile } }
     let(:loader) { ActiveFedora::SolrInstanceLoader.new(Foo, obj.id, doc) }
-    it "should load the object without trouble" do
-      expect(loader.object).to be_instance_of Foo
+    context "when the json has extra values in it" do
+      let(:profile) { { "foo"=>["baz"], "bar"=>"quix", "title"=>"My Title", "extra_value"=>"Bonus values!"}.to_json }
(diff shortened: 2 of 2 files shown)


lib/active_fedora/loadable_from_json.rb
```diff
@@ -105,6 +105,30 @@ def reflection(predicate)
       end
     end
 
+    # Adapt attributes read from Solr to possible minor changes in data model
+    # since the attributes were saved.
+    # @param attrs [Hash] attributes read from Solr
+    # @return [Hash] the adapted attributes
+    def adapt_attributes(attrs)
+      new_attributes = {}
+      self.class.attribute_names.each do |a|
```
jcoyne: Can you set this `a` variable to something more descriptive? How about `attribute_name`? Also consider using `each_with_object({}) do |attribute_name, new_attributes|` here. Then you can remove the previous line and the last line of the method.

lib/active_fedora/loadable_from_json.rb
```diff
@@ -105,6 +105,30 @@ def reflection(predicate)
...
+        reflection = self.class.reflect_on_property a
+        new_attributes[a] = if !reflection
```
mjgiarlo: Are lines 116-128 all for setting `new_attributes[a]`? I'd consider either changing the indent to make that obvious or pulling this into a private method of its own for cleaner/leaner separation.

lib/active_fedora/loadable_from_json.rb
```diff
@@ -105,6 +105,30 @@ def reflection(predicate)
...
+          attrs[a] # if this isn't a property, copy value verbatim
+        else
+          multiple=reflection.multiple?
```
mjgiarlo: Add spaces around the `=`.

lib/active_fedora/loadable_from_json.rb
mjgiarlo: :+1:

lib/active_fedora/loadable_from_json.rb
```diff
@@ -105,6 +105,30 @@ def reflection(predicate)
...
     end
 
+    # Adapt attributes read from Solr to possible minor changes in data model
+    # since the attributes were saved.
+    # @param attrs [Hash] attributes read from Solr
+    # @return [Hash] the adapted attributes
+    def adapt_attributes(attrs)
+      new_attributes = {}
+      self.class.attribute_names.each do |a|
+        reflection = self.class.reflect_on_property a
```
mjgiarlo: Could you use parentheses to wrap args in method calls?

lib/active_fedora/loadable_from_json.rb
```diff
@@ -105,6 +105,30 @@ def reflection(predicate)
...
+        new_attributes[a] = if !reflection
+          attrs[a] # if this isn't a property, copy value verbatim
+        else
+          multiple=reflection.multiple?
+          if !attrs.key? a
+            # value is missing in attrs, add [] or nil
+            multiple ? [] : nil
+          else
+            # convert an existing scalar to an array if needed
+            (multiple && !attrs[a].is_a?(Array)) ? Array(attrs[a]) : attrs[a]
```
mjgiarlo: Can this line be simplified to:

```ruby
multiple ? Array.wrap(attrs[a]) : attrs[a]
```

ojlyytinen:
I made the suggested code formatting changes and moved the big if structure into its own private method. And adapt_attributes doesn't really need to be public either so made that private too. And finally rebased.

mjgiarlo:
:+1:

mjgiarlo:
This would benefit from a couple more eyes, too.  @jcoyne @terrellt others?

PR #876, opened 2015-08-14
Stop using InboundRelationConnection

### LINKED ISSUE #875: Revert "Requst Inbound Relations"
This commit: https://github.com/projecthydra/active_fedora/commit/025a0fc3c00d764d87ba924d95699d2fbc1b32d9

When getting inbound relations, the git request are from 3-5x slower than without. Furthermore the graphs are bigger, which means parsing takes longer. 

This will probably break IndirectContainers (and maybe activefedora-aggregations).
https://github.com/projecthydra/active_fedora/issues/793

Ideally these requests could be turned on as needed.

jcoyne:
The InboundRelation prefer header was causing a 3-5x performance
penalty. We can get the same info from Solr anyway.
Reverts #796
Fixes #875

### DIFF
--- a/lib/active_fedora/associations/contained_finder.rb
+++ b/lib/active_fedora/associations/contained_finder.rb
@@ -1,17 +1,20 @@
 module ActiveFedora::Associations
-  ## 
+  ##
   # Finds the objects which associate with a given record and are contained
   # within the given container. Uses #repository to find the objects.
   class ContainedFinder
-    attr_reader :container, :repository
+    attr_reader :container, :repository, :proxy_class
     delegate :contained_ids, to: :container
     # @param [#contained_ids] container a container that records are stored
     #   under.
     # @param [#translate_uri_to_id, #find] repository a repository to build
     #   objects from.
-    def initialize(container:, repository:)
+    # @param [ActiveFedora::Base] proxy_class class that represents an
+    #   ore:Proxy
+    def initialize(container:, repository:, proxy_class:)
       @container = container
       @repository = repository
+      @proxy_class = proxy_class
     end
 
     # @param [ActiveFedora::Base] record a record which you want to find the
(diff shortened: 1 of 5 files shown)


jcoyne:
InboundRelationConnection is still useful, so I haven't removed it. We are using it for alexandria-v2 (UCSB) like so:
```ruby
    conn = ActiveFedora::InboundRelationConnection.new(ActiveFedora.fedora.connection)
    res = Ldp::Resource::RdfSource.new conn, record.uri
    res.graph.query(object: record.uri).map { |statement| ActiveFedora::Base.uri_to_id(statement.subject) }
```

jcoyne:
I tested activefedora-aggregation with this ref and it doesn't seem to cause any failures.

spec/integration/indirect_container_spec.rb
```diff
@@ -154,10 +154,9 @@ class FooHistory < ActiveFedora::Base
             it "has two related_objects" do
               expect(reloaded.related_objects).to eq [file, file2]
             end
-            it "has inbound triples" do
-              statement = file.reload.resource.query(predicate: ::RDF::URI.new('http://www.openarchives.org/ore/terms/proxyFor')).to_a.first
 
-              expect(statement.object).to eq file.resource.rdf_subject
+            it "has inbound triples" do
+              expect(Proxy.where(proxyFor_ssim: file.id).to_a).not_to be_empty
```
terrellt: This seems like a weird test for here. Is there an instance method that does the same thing?

lib/active_fedora/associations/contained_finder.rb
```diff
@@ -1,17 +1,18 @@
...
   class ContainedFinder
-    attr_reader :container, :repository
+    attr_reader :container, :repository, :proxy_class
     delegate :contained_ids, to: :container
     # @param [#contained_ids] container a container that records are stored
     #   under.
     # @param [#translate_uri_to_id, #find] repository a repository to build
     #   objects from.
-    def initialize(container:, repository:)
+    def initialize(container:, repository:, proxy_class:)
```
terrellt: Need `@param` documentation for proxy_class.

spec/integration/indirect_container_spec.rb
jcoyne: Honestly, I would probably just remove this whole block, but I felt bad doing that. I'm not sure there really are inbound triples now. Any suggestions?

terrellt: I'd remove it.

PR #879, opened 2015-08-21
Fix issues with checking the _destroy flag in nested_attributes

ojlyytinen:
Fixes two related issues with _destroy flag in nested_attributes.

call_reject_if used to immediately return false if has_destroy_flag? returned true, even if the _destroy flag wouldn't actually destroy the record. Thus you could circumvent the reject_if check in some cases. This would happen for new records (id is blank) or if allow_destroy is not set in the options. Now reject_if call is skipped only if _destroy flag is set and allow_destroy is set in options.

This still leaves the issue of having a blank id (i.e. a new record), _destroy flag being set and allow_destroy set in the options. This used to create the new record and not do anything with the _destroy flag. What's more the reject_if checks were skipped. I changed the behaviour when both id is left blank and _destroy flag is set. Instead of creating a new record it will simply skip the whole thing doing nothing with it. I think this is also the more logical behaviour, it's as if a new record was created and then immediately destroyed.

### DIFF
--- a/lib/active_fedora/nested_attributes.rb
+++ b/lib/active_fedora/nested_attributes.rb
@@ -120,6 +120,7 @@ module ActiveFedora
 
     def assign_nested_attributes_for_collection_association(association_name, attributes_collection)
       options = nested_attributes_options[association_name]
+      allow_destroy = options[:allow_destroy]
 
       if options[:limit] && attributes_collection.size > options[:limit]
         raise TooManyRecords, "Maximum #{options[:limit]} records are allowed. Got #{attributes_collection.size} records instead."
@@ -148,13 +149,12 @@ module ActiveFedora
         attributes = attributes.with_indifferent_access
 
         if attributes['id'].blank?
-          association.build(attributes.except(*UNASSIGNABLE_KEYS)) unless call_reject_if(association_name, attributes)
-
+          association.build(attributes.except(*UNASSIGNABLE_KEYS)) unless call_reject_if(association_name, attributes, allow_destroy)
         elsif existing_record = existing_records.detect { |record| record.id.to_s == attributes['id'].to_s }
-          association.send(:add_record_to_target_with_callbacks, existing_record) if !association.loaded? && !call_reject_if(association_name, attributes)
+          association.send(:add_record_to_target_with_callbacks, existing_record) if !association.loaded? && !call_reject_if(association_name, attributes, allow_destroy)
 
-          if !call_reject_if(association_name, attributes)
-            assign_to_or_mark_for_destruction(existing_record, attributes, options[:allow_destroy])
+          if !call_reject_if(association_name, attributes, allow_destroy)
+            assign_to_or_mark_for_destruction(existing_record, attributes, allow_destroy)
           end
 
         else
@@ -181,8 +181,8 @@ module ActiveFedora
       raise ObjectNotFoundError, "Couldn't find #{reflection.klass.name} with ID=#{record_id} for #{self.class.name} with ID=#{id}"
     end
 
-    def call_reject_if(association_name, attributes)
-      return false if has_destroy_flag?(attributes)
+    def call_reject_if(association_name, attributes, allow_destroy)
+      return false if has_destroy_flag?(attributes) && allow_destroy && attributes['id'].present?
       case callback = self.nested_attributes_options[association_name][:reject_if]
       when Symbol
         method(callback).arity == 0 ? send(callback) : send(callback, attributes)
@@ -193,5 +193,3 @@ module ActiveFedora
 
   end
 end
-
-  
--- a/spec/integration/nested_attribute_spec.rb
+++ b/spec/integration/nested_attribute_spec.rb
@@ -29,6 +29,11 @@ describe "NestedAttribute behavior" do
       accepts_nested_attributes_for :bars, reject_if: proc { |attributes| attributes['uno'].blank? }
     end
 
+    # class used in test for reject_if: with proc object and destroy flag set.
+    class CarProcDestroy < Car
+      accepts_nested_attributes_for :bars, allow_destroy: true, reject_if: proc { |attributes| attributes['uno'].blank? }
+    end
+
     # class used in test for reject_if: with method name as symbol
     class CarSymbol < Car
       accepts_nested_attributes_for :bars, reject_if: :uno_blank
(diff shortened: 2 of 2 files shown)


lib/active_fedora/nested_attributes.rb
```diff
@@ -148,13 +149,14 @@ def assign_nested_attributes_for_collection_association(association_name, attrib
         attributes = attributes.with_indifferent_access
 
         if attributes['id'].blank?
-          association.build(attributes.except(*UNASSIGNABLE_KEYS)) unless call_reject_if(association_name, attributes)
-
+          if !has_destroy_flag?(attributes) || !allow_destroy
```
mjgiarlo: The code here is very complex and I'm less familiar with it, so it is difficult for me to independently review the change. What happens if this added `if` block isn't evaluated? Is the association silently not built?  Hoping someone with more knowledge of this class can chime in. (ping @jcoyne)

ojlyytinen: Can't say I'm entirely familiar with this either. But my reasoning was that the association.build line has an unless clause already so I assumed it'd be safe to just skip that line if for whatever reason we don't want to create the association.

mjgiarlo:
Comments on this PR, @jcoyne?

lib/active_fedora/nested_attributes.rb
jcoyne: Would you please refactor this to `unless has_destroy_flag?(attributes) && allow_destroy`

jcoyne: Why would we even need this logic?  If they didn't provide an ID, whether or not the provided a destroy flag is irrelevant. We can't destroy anything because we can't identify it.

lib/active_fedora/nested_attributes.rb
```diff
@@ -181,8 +183,8 @@ def raise_nested_attributes_record_not_found(association_name, record_id)
       raise ObjectNotFoundError, "Couldn't find #{reflection.klass.name} with ID=#{record_id} for #{self.class.name} with ID=#{id}"
     end
 
-    def call_reject_if(association_name, attributes)
-      return false if has_destroy_flag?(attributes)
+    def call_reject_if(association_name, attributes, allow_destroy)
```
jcoyne: If they pass destroy, we should not reject it here. We check for that case in `assign_to_or_mark_for_destruction`

spec/integration/nested_attribute_spec.rb
```diff
@@ -52,80 +52,98 @@ class CarWithLimit < Car
...
-    expect {
-      @car.attributes = {:bars_attributes=>[{}]}
-    }.to_not raise_exception
-
-    expect {
-      @car.attributes = {:bars_attributes=>[{}, {}]}
-    }.to raise_exception(ActiveFedora::NestedAttributes::TooManyRecords)
+  describe "allow_destroy: true" do
+    let(:car_class) { Car }
+    it "should not create a new record if _destroy is set" do
```
jcoyne: I disagree with this. If you set _destroy and you don't set an id, it should ignore the _destroy flag and go on creating as usual.

ojlyytinen:
Before this pull request, if a user doesn't provide ID but does set the destroy flag then a new record will be created and the reject_if is not called. So the user could enter records that would otherwise be rejected by the reject_if. That's the main problem I'm trying to fix.

What to do if the user tries to create a new record (blank ID) and also sets the destroy flag is more a matter of opinion. I felt that it would be more logical if a new record won't be created in this case. It's as if a new record was created and then immediately destroyed. 

This came up in some javascript code dealing with nested attributes. Normally when a user removes a nested attribute value the javascript just hides the value and sets a hidden destroyed flag for it in the form. But if it was a newly created value without an id then this won't work, the value will still be created. I thought this was surprising behaviour, but certainly this can easily be fixed in the javascript code too. If you disagree, I can revert this behaviour back to what it was before. But I think the possibility of the user skipping reject_if checks should still be fixed.

lib/active_fedora/nested_attributes.rb
ojlyytinen: This doesn't reject the record. It only skips the reject_if check, as in reject_if returning false so it won't be rejected. The problem here was that reject_if check could be skipped by setting the destroy flag even if the destroy flag wouldn't actually destroy the record.

jcoyne:
@ojlyytinen Yes, I see your point about "if a user doesn't provide ID but does set the destroy flag then a new record will be created and the reject_if is not called." We definitely need to fix that.  I'm having a look at how rails currently solves that issue and how we are differing from them in that instance.  I want to keep this code as close to ActiveRecord as possible to make backporting easier.

jcoyne:
@ojlyytinen In your PR, I don't see a test that describes the bug you've described to me above.  The only two tests that fail are these:
https://github.com/projecthydra/active_fedora/pull/879/files#diff-ef6b350aaffb78442222fd37752d8f81R104
https://github.com/projecthydra/active_fedora/pull/879/files#diff-ef6b350aaffb78442222fd37752d8f81R119

The former has an id and the later doesn't have a proc.

I've pulled your other changes to the test into this PR: https://github.com/projecthydra/active_fedora/pull/893  starting from there, can you write a test that fails when the situation you presented is encountered?  Then change the code so that the test passes.

lib/active_fedora/nested_attributes.rb
jcoyne: Right, I think we should pull this code in: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/nested_attributes.rb#L483

ojlyytinen:
Sure, I'll do these changes Monday. In the case of id being blank and destroy flag being set, would you like it to behave as it used to and create the record? In that case I'll need to add a test to make sure that the reject_if check is still being ran. The way I have this PR now the record would not be created in any case so there was no point in calling reject_if or testing for it.

ojlyytinen:
I've rebased this on #893 and changed the behaviour a bit. The spec has three new tests, one of which passes without any changes to code.  The passing test is when you have a blank id (that is a new record) and destroy flag set, it'll still create the record. That's what always happened but there's a test for it now if that's the intended behaviour for this case. In the previous version of this pull request I had changed this not to create the record but I reverted that now back to what it did before.

The two other tests will fail (without the implementation changes) and they both are about reject_if check not being called when destroy flag is set.

There are also six other failing tests. But those same tests seem to be failing even in the master branch if I run them so I think they're unrelated. Not sure how they passed in the commits that are now merged in master or in #893.

jcoyne:
This has been encorporated into #900. Thanks @ojlyytinen !

PR #884, opened 2015-08-28
Using premis:hasMessageDigest for checksum

### LINKED ISSUE #883: AF::File::Attributes should find digests atop Fedora 4.3.0
Fedora 4.3.0 was included in the latest commit of hydra-jetty, and 4.3.0 has [changes](https://wiki.duraspace.org/display/FF/Fedora+4.3.0+Release+Notes) to three Fedora-managed properties:

> This release is 4.3.0 instead of 4.2.1 because there are two non-backwards compatible changes: Several properties (mimetype, filename, digest, and cnd update) no longer use the fedora namespace.

`AF::File::Attributes` expects the digest property to be `Fcrepo4.digest` ([here](https://github.com/projecthydra/active_fedora/blob/d6784e23a7ad324e2d03a23c7c063fc888b89398/lib/active_fedora/file/attributes.rb#L18)) but that is no longer the case; now it's `PREMIS.hasMessageDigest`. (Also, not sure if these are elsewhere in the AF codebase but `Fcrepo4.mimeType` has changed to `EBUCore.hasMimeType` and `PREMIS.hasOriginalName` has changed to `EBUCore.filename`.)

Should AF master require a particular version of Fedora or should we make the code agnostic by, say, sniffing the version of Fedora and choosing the digest predicate based on that?

escowles:
* Using new predicate (premis:hasMessageDigest) for file checksums
* Falling back on old predicate (fedora:digest) for backwards compatability
Fixes #883

### DIFF
--- a/lib/active_fedora/file/attributes.rb
+++ b/lib/active_fedora/file/attributes.rb
@@ -15,7 +15,9 @@ module ActiveFedora::File::Attributes
   end
 
   def digest
-    response = metadata.ldp_source.graph.query(predicate: ActiveFedora::RDF::Fcrepo4.digest)
+    response = metadata.ldp_source.graph.query(predicate: RDF::Vocab::PREMIS.hasMessageDigest)
+    # fallback on old predicate for checksum
+    response = metadata.ldp_source.graph.query(predicate: ActiveFedora::RDF::Fcrepo4.digest) if response.empty?
     response.map(&:object)
   end
 
--- a/spec/unit/file_spec.rb
+++ b/spec/unit/file_spec.rb
@@ -230,6 +230,16 @@ describe ActiveFedora::File do
         expect(parent.reload.abcd.digest.first).to be_kind_of RDF::URI
       end
     end
+    context "with pre-4.3.0 predicate" do
+      before do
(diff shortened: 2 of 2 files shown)


mjgiarlo:
:+1: 

@jcoyne ?

lib/active_fedora/rdf/premis.rb
```diff
@@ -0,0 +1,6 @@
+require 'rdf'
+module ActiveFedora::RDF
+  class Premis < RDF::StrictVocabulary("http://www.loc.gov/premis/rdf/v1#")
```
jcoyne: Can we use RDF::Vocab::Premis.hasMessageDigest (https://github.com/ruby-rdf/rdf-vocab/blob/develop/lib/rdf/vocab/premis.rb#L742) rather than defining our own vocabulary?

mjgiarlo: I'm :ok: with pulling rdf-vocab in and using it.

escowles:
I've updated the PR to use RDF::Vocab::PREMIS.  Related, I've also opened #886 for also using RDF::Vocab::Fcrepo4 instead of the local fcrepo4 vocab.

PR #887, opened 2015-08-29
Using Fcrepo4 and LDP from RDF::Vocab instead of local versions

### LINKED ISSUE #886: Should we use RDF::Vocab::Fcrepo4 instead of a local vocabulary?
RDF::Vocab has a Fcrepo4 vocabulary: https://github.com/ruby-rdf/rdf-vocab/blob/develop/lib/rdf/vocab/fcrepo4.rb -- switch to using it instead of ActiveFedora::RDF::Fcrepo4.

escowles:
Fixes #886


mjgiarlo:
:clap:

PR #899, opened 2015-09-24
Don't use Fcrepo digest predicate from rdf-vocab

jcoyne:
Fcrepo4.digest was used by Fedora < 4.2, but it was removed
from the 2015-07-24 version of the fedora 4 ontology
http://fedora.info/definitions/v4/2015/07/24/repository
and from rdf-vocab in version 0.8.5

### DIFF
--- a/lib/active_fedora/file/attributes.rb
+++ b/lib/active_fedora/file/attributes.rb
@@ -17,7 +17,7 @@ module ActiveFedora::File::Attributes
   def digest
     response = metadata.ldp_source.graph.query(predicate: RDF::Vocab::PREMIS.hasMessageDigest)
     # fallback on old predicate for checksum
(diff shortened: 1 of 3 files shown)


jcoyne:
See https://jira.duraspace.org/browse/FCREPO-1746

acoburn:
It seems a little dangerous to add non-existent properties to a vocabulary like this. Especially as I wouldn't expect those properties to be re-added to any future version of the fedora ontology. At the very least, you may want to add some explicit WARN (logging) output to encourage clients to not use or expect those properties.

jcoyne:
@acoburn They aren't non-existant if you are using Fcrepo 4.2  This is intended to support Fcrepo < 4.3

acoburn:
I'm :+1: on merging this as-is, given that this is designed to support fedora 4.2

PR #901, opened 2015-09-28
Allow for sub RDF Sources via contains

terrellt:
I'll squash this when it's okay - for now it shows a refactor route.

### DIFF
--- a/lib/active_fedora/association_hash.rb
+++ b/lib/active_fedora/association_hash.rb
@@ -1,0 +1,116 @@
+module ActiveFedora
+  ##
+  # Used as an access method for associations on a model, given some
+  # reflections.
+  class AssociationHash
+    def initialize (model, reflections)
+      @base = model
+      @reflections = reflections
+    end
+
+    def [] (name)
+      association(name).reader if association(name)
+    end
+
+    def []= (name, object)
+      association(name).writer(object) if association(name)
+    end
+
+    def association(name)
+      # Check to see if the key exists before casting to a symbol, because symbols
+      # are not garbage collected in earlier versions of Ruby
+      @base.association(name.to_sym) if key?(name)
+    end
+
+    def reflections
+      @reflections
+    end
+
+    def merge(other_hash)
+      Merged.new(self, other_hash)
+    end
+
+    def each
+      keys.each do |k|
+        yield k, self[k]
+      end
+    end
+
+    def keys
+      reflections.keys
+    end
+
+    # Check that the key exists with indifferent access (symbol or string) in a
+    # manner that avoids generating extra symbols. Symbols are not garbage collected
+    # in earlier versions of ruby.
+    def key?(key)
+      keys.include?(key) || keys.map(&:to_s).include?(key)
+    end
+
+    def values
+      keys.map { |k| self[k] }
+    end
+
+    def has_key?(key)
+      keys.include?(key)
+    end
+
+    def size
+      keys.size
+    end
+
+    def empty?
+      reflections.empty?
+    end
+
+    def each_value
+      keys.each do |k|
+        yield self[k]
+      end
+    end
+
+    def changed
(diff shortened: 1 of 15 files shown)


terrellt:
A side-note - this may get us close to a fix for #858

lib/active_fedora/association_hash.rb
```diff
@@ -0,0 +1,89 @@
+require 'forwardable'
+
+module ActiveFedora
+  class AssociationHash
+    extend Forwardable
```
jeremyf: Why does this extend Forwardable?

terrellt: No idea. As far as I can tell it doesn't use it - was brought over from FilesHash. I could remove it, but I'm not sure what the project's policy on refactors + feature requests in one PR is.

mjgiarlo: Unless @jcoyne voices an objection, I'd say dump it. It's a new file, so meh.

jcoyne: Yes, dump it.

lib/active_fedora/associations/basic_contains_association.rb
```diff
@@ -7,13 +7,21 @@ def reader(force_reload = false)
       end
 
       def find_target
-        reflection.build_association(target_uri) do |record|
+        find_or_initialize_target do |record|
           configure_datastream(record) if reflection.options[:block]
         end
       end
 
+      def find_or_initialize_target(&block)
```
jcoyne: Should this be private?

lib/active_fedora/persistence.rb
```diff
@@ -146,7 +146,9 @@ def create_record(options = {})
       serialize_attached_files
       @ldp_source = @ldp_source.create
       assign_uri_to_attached_files
+      assign_uri_to_child_resources
```
jcoyne: Is a child resource a hash-code resource?

jcoyne:
Do we really want to make a clear distinction between attached files and child_resources? I'd prefer we handle them in the same way if possible.

lib/active_fedora/persistence.rb
terrellt: No. Child resources are like /obj1 is an RDFSource and /obj1/obj2 is also a RDFSource. "contained_rdf_source" may be more appropriate.

terrellt:
@jcoyne Can you take another look? Did some work to combine handling files and nested rdf sources.

terrellt:
Rebased.

lib/active_fedora/associations/id_composite.rb
```diff
@@ -20,7 +20,7 @@ def each
     private
 
     def convert(id)
-      if id.start_with?("http")
```
jcoyne: Please add YARD documentation on the parameter please

lib/active_fedora/core.rb
```diff
@@ -37,6 +37,12 @@ def initialize(attributes_or_id = nil, &block)
       run_callbacks :initialize
     end
 
+    def uri=(uri)
```
jcoyne: Please add YARD doc about the parameter type(s)

lib/active_fedora/core.rb
```diff
@@ -37,6 +37,12 @@ def initialize(attributes_or_id = nil, &block)
...
+      unless persisted?
```
jcoyne: Perhaps raise an error when calling this on a persisted object.

lib/active_fedora/associations/id_composite.rb
terrellt: For the private method?

lib/active_fedora/core.rb
terrellt: Recommendation on the exception class?

PR #906, opened 2015-10-12
Fixed syntax error in raise statement

dchandekstark:


mjgiarlo:
@dchandekstark @awead @jcoyne does this imply we don't have test coverage of the `NotImplementedError`?

PR #907, opened 2015-10-12
AssociationHash - alias `#include?` to `#key?`

dchandekstark:


grosscol:
I have no feelings on this one way or another.  What use case brought this up?

dchandekstark:
@grosscol In pre-AF9, ActiveFedora::Base#datastreams returns a DatastreamHash (e.g., https://github.com/projecthydra/active_fedora/blob/v7.1.2/lib/active_fedora/datastream_hash.rb) which supports Hash instance methods (which includes `include?`).  In AF9, this becomes a FilesHash, which is an AssociationHash.  The continuity is nice, plus the names suggest that one might expect that method of Hash to work, and there seems to be no reason not to.

jcoyne:
It looks like I removed the delegators here: https://github.com/projecthydra/active_fedora/commit/ff7da65dc0c1679fc7cb61e275996212627e8535  At that commit we switched from wrapping a hash, to an API that mimics the Hash api, so it isn't complete.  Feel free to add missing methods as you need them.

PR #911, opened 2015-10-15
Support both Fedora and Premis predicates, supports Fedora 4.4

awead:
The fixity predicates have changed. We'll need to release a new version of AF once this is merged, but it requires an update to hydra-jetty.

### DIFF
--- a/lib/active_fedora/fixity_service.rb
+++ b/lib/active_fedora/fixity_service.rb
@@ -11,24 +11,33 @@ module ActiveFedora
     end
 
     # Executes a fixity check on Fedora and saves the Faraday::Response.
-    # Returns true when the fixity check was successfully.
+    # @return true or false
     def check
       @response = get_fixity_response_from_fedora
-      status.match("SUCCESS") ? true : false
+      status.include?(success)
     end
 
     def status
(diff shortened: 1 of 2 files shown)


awead:
See projecthydra/hydra-jetty#39

awead:
Is it worth modifying FixityService to support checks made to all versions of Fedora 4? This would mean querying for three difference predicates:

* ::RDF::Vocab::Fcrepo4.status
* ::RDF::URI("http://fedora.info/definitions/v4/repository#status")
* and the current Premis one

thoughts? @jcoyne @terrellt @mjgiarlo

jcoyne:
I think we should probably attempt to support at least 4.4.0 and 4.3.0, so we should probably query for both URIs.  I think there are only 2 though. Aren't the first two bullet points the same predicate? It's just that it's no longer in the most recent version of rdf-vocab, which is why we moved it to a hard coded uri.

awead:
@jcoyne yeah, that looks right. I'll update the PR.

awead:
@jcoyne I tested it locally under Fedora 4.3 and 4.4, and all is well. Travis should be testing it using 4.4, now that hydra-jetty is updated.

Issue #870, opened 2015-08-03
solr loads only one nested attribute


hackmastera:
Here is a test showing the problem:
https://github.com/chemheritage/active_fedora/tree/solr-load-bug

hackmastera:
@cam156 @awead @hectorcorrea - according to @jcoyne you all know everything about active fedora and may be able to help me get to the root of this.

awead:
@hackmastera I don't have time to look into this at the moment, but as a workaround, I would suggest creating your own solr indexer for the classes that have nested attributes. Then you could either index ids directly, or even index the properties of the nested attributes. That's might a better option anyway, because you'd have all the terms you'd need in one solr document.

Another option would be to do solr joins, but that's @cam156 territory.

hackmastera:
See also this hydra-tech list message:

https://groups.google.com/forum/#!topic/hydra-tech/bybriXid38Y

I'm going to try changing my model as suggested by that poster. That doesn't invalidate the bug, though.

hackmastera:
Thanks, @awead! It doesn't look like the indexing itself is the problem; both IDs are in solr when I look at it directly. It's definitely an issue with the retrieval.

I take your point, though, that indexing the attributes themselves would probably be useful, as well.

awead:
Ah, ok, I misunderstood the problem. The issue is likely with an incomplete object profile. Take a look at:

https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/solr_instance_loader.rb

It's loading the object based on the object_profile_ssm field in the solr document. Check and see if each nested attribute is there. I've never looked into this, so I'm not sure what you should expect.

hackmastera:
@awead, yes that's the class the test is written against. when you say "Check and see if each nested attribute is there" do you mean in my solr index itself?

ojlyytinen:
When I ran into this (presumably the same) issue earlier, I spent quite some time tracing it through Active Fedora. It's been a while but from what I remember, everything is stored correctly in Solr, but the loading overwrites values and leaves only the last one.

From my earlier post to Hydra-Tech list (linked above). When the execution gets to ActiveFedora::Associations::RDF.replace all the author_ids are still intact. But there it loops over each of them and goes to ActiveFedora::LoadableFromJson::SolrBackedResource.insert, one id at a time. That then goes to the set_value in the same file which just replaces any existing value rather than adding them all to a list, thus only the last of the author_ids is set in the end.

[...]

PR #918, opened 2015-10-20
Append val to solr array instead of replacing when appropriate.

hackmastera:
Fixes projecthydra/active_fedora#870.

### DIFF
--- a/lib/active_fedora/loadable_from_json.rb
+++ b/lib/active_fedora/loadable_from_json.rb
@@ -97,7 +97,12 @@ module ActiveFedora
       # @param [Array] vals an array of 3 elements (subject, predicate, object) to insert
       def insert(vals)
         _, pred, val = vals
-        set_value(reflection(pred), [val])
+        k = reflection(pred)
+        if @hash[k].is_a?(Array)
+          set_value(k, @hash[k] << val)
+        else
+          set_value(k, [val])
+        end
       end
 
       def reflection(predicate)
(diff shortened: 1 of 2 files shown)


lib/active_fedora/loadable_from_json.rb
```diff
@@ -45,7 +45,11 @@ def to_s
       # It is expected that the singular filter gets applied after fetching the value from this
       # resource, so cast everything back to an array.
       def set_value(k, v)
-        @hash[k] = Array(v)
+        if (!@hash[k].nil? && @hash[k].is_a?(Array))
```
jcoyne: No need to trap for nil here:
```
nil.is_a? Array
=> false
```

jcoyne: I'm not sure this refactor changes anything.  Where is the difference?

jcoyne: Oh, I see it's appending vs replacing. Why would we append in a method called "set_value"?

hackmastera:
@jcoyne thank you for your comments. How does this look?

mjgiarlo:
Looks clean to me but this is @jcoyne's wheelhouse.

spec/integration/solr_instance_loader_spec.rb
```diff
@@ -43,6 +57,14 @@ class Foo < ActiveFedora::Base
       it "should not be mutable" do
         expect { subject.title = 'Foo' }.to raise_error ActiveFedora::ReadOnlyRecord
       end
+
+      it "loads the correct number of nested attributes", focus: true do
```
jcoyne: why is `focus: true` set here?

spec/integration/solr_instance_loader_spec.rb
```diff
@@ -43,6 +57,14 @@ class Foo < ActiveFedora::Base
...
+        expect(subject.date_ids.count).to eq 2
+      end
+
+      it "loads the correct number of regular attributes", focus: true do
```
jcoyne: why `focus`?

hackmastera: Ugh, good catch! ha, AF's rspec wasn't even set up for this so I didn't notice it was there.

PR #919, opened 2015-10-20
Adds ActiveFedora::Checksum class to encapsulate file digest info.

dchandekstark:
Adds `#checksum` method to ActiveFedora::File.

### DIFF
--- a/lib/active_fedora/file.rb
+++ b/lib/active_fedora/file.rb
@@ -134,6 +134,10 @@ module ActiveFedora
       @metadata ||= ActiveFedora::WithMetadata::MetadataNode.new(self)
     end
 
+    def checksum
+      ActiveFedora::Checksum.new(self)
+    end
+
     def content_changed?
       return true if new_record? and !local_or_remote_content(false).blank?
       local_or_remote_content(false) != @ds_content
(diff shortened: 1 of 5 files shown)


jcoyne:
:+1: @awead care to have a look?

lib/active_fedora/file.rb
```diff
@@ -134,6 +134,10 @@ def metadata
       @metadata ||= ActiveFedora::WithMetadata::MetadataNode.new(self)
     end
 
+    def checksum
+      ActiveFedora::Checksum.new(self)
```
awead: Maybe:
```
@checksum ||= ActiveFedora::Checksum.new(self)
```
and then set `@checksum = nil` under #refresh.

dchandekstark: Personally I don't think the operation cost warrants the complexity of caching.  Or if it does, perhaps it belongs on `ActiveFedora::File::Attributes#digest`.  @jcoyne

awead: Yeah, I'd buy that.

dchandekstark: OK, well maybe we merge this and address any caching separately if anybody thinks that's important?

PR #921, opened 2015-10-22
Load create_date and modified_date from Solr.

### LINKED ISSUE #905: Create_date and modified_date raise errors with instances loaded from Solr
If you load an instance from Solr and try to access create_date or modified_date, you get

```
     NoMethodError:
       undefined method `first' for nil:NilClass
     # ./lib/active_fedora/fedora_attributes.rb:18:in `create_date'
     # ./spec/integration/solr_instance_loader_spec.rb:130:in `block (4 levels) in <top (required)>'
```

I have this addition in spec/integration/solr_instance_loader_spec.rb

```ruby
  describe "loading system properties" do
    let(:obj_solr) { Foo.load_instance_from_solr(obj.id) }
    describe "create_date" do
      it "loads it from solr" do
        expect(obj.create_date).to be_present
        expect(obj_solr).to be_present
        expect(obj_solr.create_date).to be_present # This is line 130
        expect(obj_solr.create_date).to be_a DateTime
      end
    end

    describe "modified_date" do
      it "loads modified_date from solr" do
        expect(obj.modified_date).to be_present
        expect(obj_solr).to be_present
        expect(obj_solr.modified_date).to be_present
        expect(obj_solr.modified_date).to be_a DateTime
      end
    end
  end
```

I can see a few ways to fix this issue and am quite happy to do the pull request myself. I'm just not sure what the preferred way to fix it would be. Here are a few options I thought of

1. Modify SolrInstanceLoader#profile_json. The entire solr_doc is available here and system_create_dtsi and system_modified_dtsi can be taken from there. But the return value is a string serialisation of the profile json so it's a bit messy to inject that in there. 
2. Change LoadableFromJson#init_with_json to take another optional hash which could be used to pass additional values from SolrInstanceLoader#allocate_object. Then just add another method in SolrInstanceLoader to get the dates from the solr_doc and pass them on.
3. Similar to previous but instead of an extra parameter to init_with_json give a block from SolrInstanceLoader that gets evaluated.
4. Add the dates to the profile_json in ProfileIndexingService#export or IndexingService#generate_solr_document. 

All these options except 3 would need create_date and modified_date to be added to LoadableFromJson#adapt_attributes to get them go through the attributes assignment in init_with_json.

ojlyytinen:
Fixes #905. Trying to access create_date or modified_date in objects loaded from Solr used to raise an error. They are now loaded properly from the Solr properties system_create and system_modified.

### DIFF
--- a/lib/active_fedora/solr_instance_loader.rb
+++ b/lib/active_fedora/solr_instance_loader.rb
@@ -27,7 +27,12 @@ module ActiveFedora
     private
 
       def allocate_object
-        active_fedora_class.allocate.init_with_json(profile_json)
+        active_fedora_class.allocate.init_with_json(profile_json) do |allocated_object|
+          create_key = allocated_object.indexing_service.class.create_time_solr_name
+          modified_key = allocated_object.indexing_service.class.modified_time_solr_name
+          allocated_object.resource.set_value(:create_date, DateTime.parse(solr_doc[create_key])) if solr_doc[create_key]
+          allocated_object.resource.set_value(:modified_date, DateTime.parse(solr_doc[modified_key])) if solr_doc[modified_key]
+        end
       end
 
       def solr_doc
--- a/lib/active_fedora/indexing_service.rb
+++ b/lib/active_fedora/indexing_service.rb
@@ -19,6 +19,14 @@ module ActiveFedora
       @profile_solr_name ||= ActiveFedora::SolrQueryBuilder.solr_name("object_profile", :displayable)
     end
 
+    def self.create_time_solr_name
+      @create_time_solr_name ||= ActiveFedora::SolrQueryBuilder.solr_name('system_create', :stored_sortable, type: :date)
+    end
+
+    def self.modified_time_solr_name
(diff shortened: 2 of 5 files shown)


awead:
:+1:

lib/active_fedora/solr_instance_loader.rb
```diff
@@ -27,7 +27,12 @@ def object
     private
 
     def allocate_object
-      active_fedora_class.allocate.init_with_json(profile_json)
+      active_fedora_class.allocate.init_with_json(profile_json) do |allocated_object|
+        create_key = Solrizer.solr_name('system_create', :stored_sortable, type: :date)
```
jcoyne: This code (to get the key) is already in Querying. Can we extract it to a constant?

ojlyytinen:
I added some class methods to get the Solr names for the fields in ActiveFedora::IndexingService as that seemed the most appropriate place and it already had one for the object profile. 

I tried to change the few lines in generate_solr_document to use the new methods as well but it seems that dates really need to go through the Solr field mapper to work properly and I couldn't immediately figure out how to pass the already solrized name to Solrizer.set_field.

jcoyne:
Looks good. Please rebase and run rubocop and this can be merged.

ojlyytinen:
Rebased.

PR #922, opened 2015-10-22
Return values for inherited attributes the same way we do for properties

awead:


jcoyne:
Instead of doing it this way, can we make `obj[:type]` return a value?  Everything that is a tracked attribute should be able to be accessed via`[]`

awead:
@jcoyne I looked into that, and it looked difficult. It would involve changing the behaviors of ActiveFedora::FedoraAttributes so they would appear as delegated_attributes, but they're not really delegated like properties... right?

jcoyne:
`:type` should act just like any other property.   There is no reason we shouldn't be able to do `obj[:type]` so if you don't fix it that way, then we need to open a new ticket that says type should be fetchable that way.

awead:
@jcoyne how about this?

lib/active_fedora/attributes.rb
```diff
@@ -126,6 +126,11 @@ def system_attributes
         ['has_model', 'create_date', 'modified_date']
       end
 
+      # Attributes inherited from Fedora in ActiveFedora::FedoraAttributes
+      def inheritable_attributes
```
jcoyne: I'm not sure I understand why these are "inheritable". Can you say more?

jcoyne:
This way is much better, I guess I'm just hung up on why this is called an "inheritable_attribute".

lib/active_fedora/attributes.rb
awead: I got it from this method:
https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/fedora_attributes.rb#L25
But I'm not attached to the name. I was hoping for a way to pull the names of the accessors from the FedoraAttributes module, but didn't see a clear way to do this, so I just opted to repeating the terms here.

jcoyne: Yes, it's not the attributes that are inheritable here, it is the methods that are (similar to class_attribute) https://github.com/projecthydra/active_fedora/blob/80b830ee76ce9407058250a00f83b570a46aa7ee/lib/active_fedora/inheritable_accessors.rb#L8  How about `attributes_with_defaults` ?

awead: sure. I'll update the PR.

awead:
I changed the method name to `attributes_with_defaults`

PR #923, opened 2015-10-22
Save time zone information for DateTimes. Solr returns DateTimes rather than Strings.

ojlyytinen:
As it stands, time zone information is not stored in Fedora, instead it's just always set to Z. All time information is effectively shifted by your local time offset if you take time zones into account. This results in unexpected behaviour where timestamps for example end up in the future if your local time zone offset is positive.

```ruby
  property :date, predicate: ::RDF::DC.date do |index|
    index.type :date
  end
end

foo = Foo.create(date: [DateTime.now])
sleep 1
foo.date.first > DateTime.now # => true assuming your time offset is positive
```

There seems to be a bug in RDF::Literal::DateTime#to_s which prints out time zone as Z if you initialise the object with a DateTime rather than a String, so this should be fixed there. But a similar bug is present also in the monkey patch applied in [lib/active_fedora.rb](https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora.rb#L15-L35) which is fixed in this PR.

The PR also makes Solr-loaded documents return DateTime objects rather than Strings when indexing type has been set to date. What used to happen is that the same field which returned a DateTime object when loaded from Fedora would return a String when loaded from Solr.

### DIFF
--- a/spec/integration/date_time_properties_spec.rb
+++ b/spec/integration/date_time_properties_spec.rb
@@ -1,0 +1,40 @@
+require 'spec_helper'
+
+describe ActiveFedora::Base do
+  before do
+    class Foo < ActiveFedora::Base
+      # Date attributes are recognized by having index.type :Date or class_name: 'DateTime'
+      property :date, predicate: ::RDF::DC.date do |index|
+        index.type :date
+      end
+      property :single_date, multiple: false, class_name: 'DateTime', predicate: ::RDF::URI.new('http://www.example.com/single_date')
+    end
+  end
+
+  after do
+    Object.send(:remove_const, :Foo)
+  end
+
+  let(:date) { DateTime.parse("2015-10-22T10:20:03.653+01:00") }
+  let(:date2) { DateTime.parse("2015-10-22T15:34:20.323-11:00") }
+  subject { Foo.create(date: [date], single_date: date2).reload }
+
+  describe "saving and loading in Fedora" do
+    it "loads the correct time" do
+      expect(subject.date.first).to eql date
+      expect(subject.single_date).to eql date2
+    end
+  end
+
+  describe "saving and loading in Solr" do
+    let(:subject_solr) { subject.class.load_instance_from_solr(subject.id) }
+    it "uses DateTime objects" do
+      expect(subject_solr.date.first).to be_a DateTime
+      expect(subject_solr.single_date).to be_a DateTime
+    end
+    it "loads the correct time" do
+      expect(subject_solr.date.first).to eql date
+      expect(subject_solr.single_date).to eql date2
+    end
+  end
+end
--- a/lib/active_fedora.rb
+++ b/lib/active_fedora.rb
@@ -14,20 +14,21 @@ ENABLE_SOLR_UPDATES = true unless defined?(ENABLE_SOLR_UPDATES)
 
 # Monkey patching RDF::Literal::DateTime to support fractional seconds.
 # See https://github.com/projecthydra/active_fedora/issues/497
+# Also monkey patches in a fix for timezones to be stored properly.
 module RDF
   class Literal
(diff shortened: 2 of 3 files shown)


spec/integration/date_time_properties_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  after do
+    Object.send(:remove_const, :Foo)
+  end
+
+  let(:date) { DateTime.parse("2015-10-22T10:20:03.653+01:00") }
+  let(:date2) { DateTime.parse("2015-10-22T15:34:20.323-11:00") }
+  subject { Foo.create(date: [date], single_date: date2) }
+
+  describe "saving and loading in Fedora" do
```
jcoyne: Where is the "load" happening in this scenario?

spec/integration/date_time_properties_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'spec_helper'
+
+describe ActiveFedora::Base do
+  before do
+    class Foo < ActiveFedora::Base
+      property :date, predicate: ::RDF::DC.date do |index|
+        index.type :date
```
jcoyne: Is the "type" necessary?  It should be inferred from what is given to the indexer, right?

spec/integration/date_time_properties_spec.rb
ojlyytinen: It seems to reset attributes after save (or create). At least the time zone gets reset with create alone and that test fails. I could add a .reload in there though just in case.

spec/integration/date_time_properties_spec.rb
ojlyytinen: Date is stored in Solr as a String regardless of the type. When loading from Solr, I think that's the only type information available. Or is there some other way to infer the type? Does the indexer add a date suffix (_dtsi or something) in the field name if it's given a date even if the index.type isn't set? If this is the case then I guess the type could be inferred from the suffix alone.

jcoyne: I think we should look at the suffix, because there may be date fields that don't declare `index.type :date`  For example this line: https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/indexing_service.rb#L31  produces: `"system_create_dtsi"=>"2015-10-23T14:06:19Z"`

spec/integration/date_time_properties_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  after do
+    Object.send(:remove_const, :Foo)
+  end
+
+  let(:date) { DateTime.parse("2015-10-22T10:20:03.653+01:00") }
+  let(:date2) { DateTime.parse("2015-10-22T15:34:20.323-11:00") }
+  subject { Foo.create(date: [date], single_date: date2) }
+
+  describe "saving and loading in Fedora" do
+    it "uses DateTime objects" do
```
jcoyne: This block doesn't seem necessary as it's not impacted by the changes you made to the code, it's covered by ActiveTriples' tests and largely covered by the block 28-31

jcoyne:
Please rebase your changes and run `rubocop`.  I suspect it will tell you to remove a bunch of `return`s at the end of methods.

ojlyytinen:
We can't really use the field name suffixes to determine the type of the field since loading only uses the object_profile which doesn't have those suffixes. As far as I can tell, the only information we have about the field is what's in ActiveFedora::Attributes::NodeConfig and the attribute value itself. I've now changed it to parse the value as a DateTime if the node config has type == :date or class_name == DateTime. When defining properties these are set with index.type :date and class_name: 'DateTime', respectively. Either will work.

Rubocop didn't mind the returns at the end of methods but I removed them anyway. And it didn't like the complexity of adapt_attribute_value so I restructured it a little bit.

PR #926, opened 2015-10-23
Create an ids_reader for indirect containers

jcoyne:
This will enable us to resolve a performance concern.
see https://github.com/projecthydra-labs/hydra-pcdm/issues/193

### DIFF
--- a/lib/active_fedora/associations/builder/indirectly_contains.rb
+++ b/lib/active_fedora/associations/builder/indirectly_contains.rb
@@ -10,6 +10,14 @@ module ActiveFedora::Associations::Builder
       reflection
     end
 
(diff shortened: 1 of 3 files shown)


jcoyne:
Please don't merge this yet. I have a few concerns.

jcoyne:
Okay, concerns addressed. Ready for review.

tpendragon:
I'm going to have to pull this locally and check. Have some concerns about what happens when there's persisted items, it's reloaded, and then you concat in memory items

mjgiarlo:
Waiting on @tpendragon's assessment.

tpendragon:
Ah, that's actually what the second test does, wasn't reading it right because of the reloaded let.

PR #933, opened 2015-10-29
Allow developers to override resource_class.

tpendragon:


jcoyne:
Looks good, but what are you going to do with this new power?

tpendragon:
> Looks good, but what are you going to do with this new power?

I've reached for it a few times for cases where an override in ActiveTriples makes sense, but I can't actually do it. I only thought about it now trying to implement the delegate query stuff locally for list_source as a way to test.

mjgiarlo:
Looks :ok: to me!

PR #936, opened 2015-11-02
Loosen dependency on rdf-rdfxml

jcoyne:


no-reply:
Is this in service of a push to `rdf` 1.99?

jcoyne:
@no-reply precisely that.

PR #955, opened 2015-11-25
Allowing access to the gone? method from outside the object

cam156:
We are running into instances where our minter is minting ids that have been deleted.  I would like access to the gone method here https://github.com/psu-stewardship/active_fedora-noid/blob/master/lib/active_fedora/noid/synchronized_minter.rb#L16

This way we will not create invalid ids and end up with an LDP::Gone exception when we try to save.


awead:
:oncoming_police_car:  busted! But otherwise, :+1:

cam156:
@awead all good now.  Darn spacing requirements...

PR #956, opened 2015-12-03
Restore versionable? behavior

awead:
Although deprecated, has_many_versions still ought to behave as it did and set self.versionable to true. This method is still used in the fedora-migrate gem as a means of identifying Fedora 3 datastreams that should have all their versions migrated.

### DIFF
--- a/lib/active_fedora/versionable.rb
+++ b/lib/active_fedora/versionable.rb
@@ -10,7 +10,7 @@ module ActiveFedora
     module ClassMethods
       def has_many_versions
         Deprecation.warn Versionable, "has_many_versions is deprecated and will be removed in ActiveFedora 11."
-        # self.versionable = true
+        self.versionable = true
       end
     end
 


lib/active_fedora/versionable.rb
```diff
@@ -10,7 +10,7 @@ module Versionable
     module ClassMethods
       def has_many_versions
         Deprecation.warn Versionable, "has_many_versions is deprecated and will be removed in ActiveFedora 11."
-        # self.versionable = true
+        self.versionable = true
```
jcoyne: Add a comment here that this is just for marking datastreams with versions for migration by fedora-migrate.

jcoyne:
I wonder if this shouldn't be a change to fedora-migrate then?

awead:
@jcoyne yes, it's actually just a change to the AF models themselves. Just replace `has_many_versions` with `self.versionable = true` and you'll get the same results. Maybe just amend the deprecation notice to say that?

PR #960, opened 2015-12-15
Add support for customizable Solr request handlers.

ojlyytinen:
Lets you configure which Solr request handlers are used for selects and updates by specifying select_path and update_path in solr.yml. This requires a patch in RSolr as well, a pull request for that is at rsolr/rsolr#116. Probably best to wait merging this until that's merged just in case something in there needs to be changed.

I've made this backwards compatible so that Active-Fedora doesn't depend on the new version of RSolr. With an older version of RSolr the custom request handlers simply will not work.

### DIFF
--- a/lib/active_fedora/solr_service.rb
+++ b/lib/active_fedora/solr_service.rb
@@ -23,11 +23,15 @@ module ActiveFedora
         Thread.current[:solr_service] = nil
       end
 
+      def select_path
+        ActiveFedora.solr_config.fetch(:select_path, 'select')
+      end
+
       def instance
         # Register Solr
 
         unless Thread.current[:solr_service]
-          register(ActiveFedora.solr_config[:url])
+          register(ActiveFedora.solr_config[:url], ActiveFedora.solr_config)
         end
 
         raise SolrNotInitialized unless Thread.current[:solr_service]
@@ -104,7 +108,7 @@ module ActiveFedora
       def query(query, args = {})
         raw = args.delete(:raw)
         args = args.merge(q: query, qt: 'standard')
-        result = SolrService.instance.conn.get('select', params: args)
+        result = SolrService.instance.conn.get(select_path, params: args)
         return result if raw
         result['response']['docs']
       end
--- a/spec/unit/finder_methods_spec.rb
+++ b/spec/unit/finder_methods_spec.rb
@@ -6,6 +6,14 @@ describe ActiveFedora::FinderMethods do
       def self.delegated_attributes
         {}
       end
+
+      def self.solr_query_handler
+        'standard'
+      end
+
+      def self.default_sort_params
+        ["system_create_dtsi asc"]
+      end
     end
   end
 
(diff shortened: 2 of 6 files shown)


awead:
@ojlyytinen can you copy in an example of what the new solr.yml is going to look like with custom update and select path? I'm having trouble visualizing this.

ojlyytinen:
Solr.yml would look like something like this. 

```ruby
development:
  url: http://localhost:8983/solr/development
  select_path: select_custom
  update_path: update_custom
```

select_path and update_path can be left out in which case it'll use the defaults.

And you'd have these defined in solrconfig.xml

```xml
  <requestHandler name="/select_custom" class="solr.SearchHandler" default="true">
     <lst name="defaults">
       ...
     </lst>
  </requestHandler>
  <requestHandler name="/update_custom" class="solr.UpdateRequestHandler" >
    ...
  </requestHandler>
```

awead:
@ojlyytinen Sorry, I forgot about this one...looks okay to me. If you want to rebase it, I can merge it.

ojlyytinen:
I think we should get the RSolr patch merged first. This one won't actually do anything until RSolr supports configurable handlers. Is there anything we can do to expedite that?

ojlyytinen:
The RSolr patch has been merged and I have also rebased this. There were a few small changes in this PR as a result of changing how RSolr handles request handlers. Functionally it should be identical though. @awead if you have time to do a review of this that would be much appreciated.

lib/active_fedora/solr_service.rb
```diff
@@ -104,7 +105,8 @@ def construct_query_for_rel(field_pairs, join_with = 'AND')
       def query(query, args = {})
         raw = args.delete(:raw)
         args = args.merge(q: query, qt: 'standard')
-        result = SolrService.instance.conn.get('select', params: args)
+        instance = SolrService.instance
+        result = instance.conn.get(instance.select_path, params: args)
```
awead: What's the difference between `@select_path` and `instance.select_path`?

spec/unit/finder_methods_spec.rb
```diff
@@ -47,4 +55,34 @@ def initialize
                              "_query_:\"{!raw f=library_id}four\"" }
     end
   end
+
+  describe "#find_in_batches" do
+    let(:docs) {
```
awead: Minor quibble: we use do/end for multiline blocks and {} for single-line blocks.

spec/unit/finder_methods_spec.rb
```diff
@@ -47,4 +55,34 @@ def initialize
...
+      double('docs').tap do |docs|
+        allow(docs).to receive(:has_next?).and_return(false)
+      end
+    }
+    let(:select_handler) { 'select' }
+    let(:connection) { double('conn') }
+    before {
```
awead: same here.

spec/unit/finder_methods_spec.rb
```diff
@@ -47,4 +55,34 @@ def initialize
...
+      expect(finder).to receive(:create_query).with('age_t' => '21').and_return('dummy query')
+      allow(ActiveFedora::SolrService.instance).to receive(:conn).and_return(connection)
+      allow(ActiveFedora::SolrService.instance).to receive(:select_path).and_return(select_handler)
+      expect(connection).to receive(:paginate) \
+        .with(1, 1000, select_handler, params: hash_including(other_opt: 'test')) \
+        .and_return('response' => { 'docs' => docs })
+    }
+    it "yields the docs" do
+      expect { |b|
```
awead: I think it's okay in this case, tho.

ojlyytinen:
I moved select_path from SolrService.instance to SolrService class level since it wasn't actually ever used in any of the instance methods. I think it's a bit cleaner that way.

I also cleaned up the block formatting issues in specs.

awead:
@ojlyytinen :clap:

PR #963, opened 2016-01-20
Updating to the latest Rubocop

awead:


jcoyne:
:+1:

.rubocop.yml
```diff
@@ -121,6 +121,8 @@ Style/PredicateName:
 Style/GuardClause:
   Exclude:
     - 'lib/active_fedora/autosave_association.rb'
+    - 'lib/active_fedora/file_configurator.rb'
```
mjgiarlo: Can you say why you added the three new line-item exclusions here rather than whipping the code into shape? Is it because there are too many violations to handle at this time?

mjgiarlo:
I have an open question about the new line-item style exclusions. I'd like to hear @awead's rationale before merging, if that's OK. (Sometimes I just add new line-item exclusions when I should probably just take ten minutes and fix 'em up, so I thought it worth asking... we can all help each other out here.)

.rubocop.yml
awead: @mjgiarlo IIRC, it wanted me to put a guard clause here:
https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/file_configurator.rb#L151:L162
at line 159, which I think was its veiled way of saying you have too many `elsif`s. I don't really have the time to pick it apart, but we could ticket it if you like.

cbeer: I'm not sure why that'd trigger a `Style/GuardClause`, but converting it to a `case` may be an easy improvement?

jcoyne: Since every case is a `return`, we can probably just omit the returns.

awead: good points. I'll try and gussy it up with a case and no returns.

awead:
@mjgiarlo :eyeglasses:

mjgiarlo:
I added a commit that removes some easy-to-remove style exclusions. I didn't rebase because I want to be sure Travis and @jcoyne are happy with them first. ;) Let's see what happens. Here's the commit: https://github.com/projecthydra/active_fedora/commit/704226d6ac75a5c3257b59376a02539ed028a9c0

awead:
Thanks, @mjgiarlo !

jcoyne:
@mjgiarlo LGTM! :+1:

awead:
@mjgiarlo feel free to squash it once travis is :green_heart:

mjgiarlo:
Pushed a new commit to deal with the breakage that was probably due to changing `and` to `&&`. That is totally yucky.

mjgiarlo:
@awead @jcoyne Build is now fixed. Squashed and pushed. :shipit: if Travis is still :green_apple: ?

awead:
:+1:

PR #966, opened 2016-01-29
Pass hash of options to index.as

awead:
Allows you to configure how we index properties that are ActiveFedora::Base objects. Instead of using only the uri as the value of the solr field, you can specify any arbitrary property you want.

@jcoyne @tpendragon I'm curious as to what your thoughts are on this approach.


mjgiarlo:
Looks clean and useful to me. :+1:

PR #967, opened 2016-01-29
Handles the case where the user passes an id to find via a hash

hectorcorrea:
This handles the problem that Jenn and Lynette reported here: https://groups.google.com/forum/#!topic/hydra-tech/72KKRPbZlms

The basic issue is that a call to `Collection.find(id: 'col-1')` behaves different than a call to `Collection.find('col-1')` Notice that the former passes the `id` as a hash.

Passing the `id` as a hash is a deprecated behavior and AF gives a deprecation warning, however, the current behavior in AF also executes the search and returns a result different than the expected one.

For example:
```
x = Collection.find(id: 'col-1')
puts x.class 
=> ActiveFedora::Relation

vs

x = Collection.find('col-1')
puts x.class 
=> Collection
```

This PR gives a more specific warning when the user issues `.find(id: "1")` and executes the correct find behavior.

### DIFF
--- a/lib/active_fedora/relation/finder_methods.rb
+++ b/lib/active_fedora/relation/finder_methods.rb
@@ -49,6 +49,10 @@ module ActiveFedora
       end
 
       if options.present?
+        if options.keys.count == 1 && options[:id]
+          Deprecation.warn(ActiveFedora::Base, "Calling .find with an id via a hash has been deprecated and will not be allowed in active-fedora 10.0. Use .find('your-id') instead")
+          return find_with_ids([options[:id]], cast)
+        end
         options = args.first unless args.empty?
         Deprecation.warn(ActiveFedora::Base, "Calling .find with a hash has been deprecated and will not be allowed in active-fedora 10.0. Use .where instead")
         options = { conditions: options }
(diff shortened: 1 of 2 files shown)


spec/unit/finder_methods_spec.rb
```diff
@@ -47,4 +47,14 @@ def initialize
                              "_query_:\"{!raw f=library_id}four\"" }
     end
   end
+
+  describe "common rookie bad call" do
```
mjgiarlo: :laughing:

jcoyne:
This would be a breaking change and thus be scheduled for AF 10, but in that same release we'll be removing this behavior.

hectorcorrea:
@jcoyne I thought we could include this in a patch or minor release, but if you think this is too big of a change for that then I am OK closing this PR.

I've updated the test so that everything passes in case you decide to merge it.

jcoyne:
I'm pretty sure that returning a list was the expected behavior when a hash was passed.  That's how we found things before `where` was introduced.

hectorcorrea:
With this PR the only thing that will change is when the user calls `find()` with a single hash of `id`, all other calls are not affected. For example: `find(id: "1")` will be affected but not `find(author:"joe", title: "xyz")`

jcoyne:
How about `find(id: ["1", "2"])` ?

PR #971, opened 2016-02-05
Generate solr 5 configs

jcoyne:


cam156:
+1 Looks like a good idea to me.

PR #973, opened 2016-02-05
Runs AF::File callbacks (fixes #972)

### LINKED ISSUE #972: AF::File callbacks are not triggered
`:create` and `:save` callbacks, e.g., are only triggered if I add this to my subclass of AF::File:

```ruby
    define_model_callbacks :update

    def create_record(*)
      run_callbacks(:create) do
        run_callbacks(:save) { super }
      end
    end

    def update_record(*)
      run_callbacks(:update) do
        run_callbacks(:save) { super }
      end
    end
```

dchandekstark:
Adds :update callback.
Callbacks were previously defined on AF::File, but not run.

### DIFF
--- a/lib/active_fedora/file.rb
+++ b/lib/active_fedora/file.rb
@@ -23,7 +23,7 @@ module ActiveFedora
 
     generate_method 'content'
 
-    define_model_callbacks :save, :create, :destroy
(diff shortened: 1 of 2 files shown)


dchandekstark:
There is a refactoring opportunity with AF::Callbacks, but I didn't want to take that on here.

PR #980, opened 2016-02-08
Don't default facet.limit

jcoyne:
When this happens, facets only show the first 10 values and don't give
any indication that there is more than 10 (the more >>) link.  This will
allow users to customize a sensible limit in their Blacklight
Configuration

### DIFF
--- a/lib/generators/active_fedora/config/solr/templates/solr/config/solrconfig.xml
+++ b/lib/generators/active_fedora/config/solr/templates/solr/config/solrconfig.xml
@@ -119,9 +119,6 @@
 
        <str name="facet">true</str>
        <str name="facet.mincount">1</str>
-       <str name="facet.limit">10</str>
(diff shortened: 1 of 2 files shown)


lib/generators/active_fedora/config/solr/templates/solr/config/solrconfig.xml
```diff
@@ -119,9 +119,6 @@
 
        <str name="facet">true</str>
        <str name="facet.mincount">1</str>
-       <str name="facet.limit">10</str>
-       <str name="facet.field">active_fedora_model_ssi</str>
```
cbeer: Did you mean to also remove these `facet.field` entries (not that I'm defending them..)?

jcoyne: Yes. I can't see any reason they're needed

PR #988, opened 2016-02-12
Share the test server method with downstream apps

jcoyne:
In your rake file add:
```ruby
 require 'active_fedora/rake_support'
```
Then you can define a task using with_test_server. For example:
```ruby
  task :ci do
    with_test_server do
      Rake::Task['spec'].invoke
    end
  end
```


awead:
:+1:

PR #993, opened 2016-02-15
Generalize the with_server method

jcoyne:
This will enable Hydra-head to provide a rake task that starts up a
development server.


awead:
:+1: Great! Will there need to be any documentation about the ENV variables, or is this a one-line rake task kind of thing?

jcoyne:
@awead I don't think so. This PR isn't adding any new variables, it's just allowing it to work with other environments. The rake task to go into Hydra-head should look like this: https://gist.github.com/jcoyne/166cefa9d917877dc2d0

PR #1002, opened 2016-03-09
Unify type to model mapping code

cbeer:
This pull request extracts two new classes from formerly class-level methods:

- `ModelClassifier` - for turning strings into class instances, and choosing the "best" match from a list
- `DefaultModelMapper` - for extracting model mapping information from resources or solr documents

This is a start at providing a flexible way to select appropriate model classes from resource data (say, from a different Solr field, or more complex logic).

### DIFF
--- a/lib/active_fedora/model.rb
+++ b/lib/active_fedora/model.rb
@@ -53,7 +53,7 @@ module ActiveFedora
         ActiveFedora::Base.logger.warn "'#{model_value}' is not a real class" if ActiveFedora::Base.logger
         return nil
       end
-      ActiveFedora.class_from_string(model_value)
+      ActiveFedora::Model.class_from_string(model_value)
     end
 
     def self.class_exists?(class_name)
--- a/lib/active_fedora/model.rb
+++ b/lib/active_fedora/model.rb
@@ -5,21 +5,6 @@ module ActiveFedora
   # This module mixes various methods into the including class,
   # much in the way ActiveRecord does.
   module Model
-    def self.best_class_from_uris(uris, default: nil)
-      best_model_match = default
-
-      uris.each do |uri|
-        model_value = from_class_uri(uri)
-        next unless model_value
-        best_model_match ||= model_value
-
-        # If there is an inheritance structure, use the most specific case.
-        best_model_match = model_value if best_model_match > model_value
-      end
-
-      best_model_match || ActiveFedora::Base
-    end
-
     # Convenience method for getting class constant based on a string
     # @example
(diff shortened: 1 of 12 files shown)


lib/active_fedora/model.rb
```diff
@@ -5,6 +5,45 @@ module ActiveFedora
   # This module mixes various methods into the including class,
   # much in the way ActiveRecord does.
   module Model
+    def self.best_class_from_uris(uris, default: nil)
```
mjgiarlo: I'm wondering if this can be tightened up a bit by removing the primitive as a default, like the following (but I'm not sure a case or two don't fall out of this one):

```ruby
def self.best_class_from_uris(uris, default: ActiveFedora::Base)
  best_model_match = default

  uris.each do |uri|
    model_value = from_class_uri(uri)
    next unless model_value

    # If there is an inheritance structure, use the most specific case.
    best_model_match = model_value if best_model_match > model_value
  end

  best_model_match
end
```

mjgiarlo: When you get out of WIP mode, it would be good to see some YARD docs for this method, even if it's just a one-line description + some `@examples`, like the method beneath it.

lib/active_fedora/query_result_builder.rb
```diff
@@ -29,20 +29,13 @@ def self.classes_from_solr_document(hit, _opts = {})
...
-        # Set as the first model in case opts[:class] was nil
-        best_model_match ||= model_value
-
-        # If there is an inheritance structure, use the most specific case.
-        best_model_match = model_value if best_model_match > model_value
-      end
+      models = Array(hit[HAS_MODEL_SOLR_FIELD])
+      best_model_match = ActiveFedora.model_mapper.best_class_from_uris(models, default: default_model)
 
       ActiveFedora::Base.logger.warn "Could not find a model for #{hit['id']}, defaulting to ActiveFedora::Base" if ActiveFedora::Base.logger && !best_model_match
```
mjgiarlo: This is true, but I wonder if this logger message ought to move into `best_class_from_uris` since that's where the defaulting is happening now.

lib/active_fedora/model.rb
cbeer: Of course; I think this method is about to disappear entirely.

lib/active_fedora/query_result_builder.rb
cbeer: Sure; can you confirm if it also makes sense to log that error from https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/relation/finder_methods.rb#L197?

PR #1009, opened 2016-03-15
Rename find_with_conditions and find_in_batches to search_with_conditions and search_in_batches

cbeer:
`find_with_conditions` and `find_in_batches` return Solr results, not class instances. Rename these methods (and provide appropriate deprecation warnings) to clarify the intent.

### DIFF
--- a/lib/active_fedora/relation/finder_methods.rb
+++ b/lib/active_fedora/relation/finder_methods.rb
@@ -87,7 +87,7 @@ module ActiveFedora
       return false unless conditions
       case conditions
       when Hash
-        find_with_conditions(conditions, rows: 1).present?
+        search_with_conditions(conditions, rows: 1).present?
       when String
         find(conditions).present?
       else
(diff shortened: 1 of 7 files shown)


lib/active_fedora/relation/finder_methods.rb
```diff
@@ -105,12 +105,18 @@ def exists?(conditions)
...
-    def find_with_conditions(conditions, opts = {})
+    def search_with_conditions(conditions, opts = {})
       # set default sort to created date ascending
       opts[:sort] = @klass.default_sort_params unless opts.include?(:sort)
       SolrService.query(create_query(conditions), opts)
     end
 
+    # @deprecated
+    def find_with_conditions(*args)
+      Deprecation.warn(ActiveFedora::Base, '.find_with_conditions is deprecated; use .search_with_conditions instead')
```
mjgiarlo: Should we add that it'll be going away in 11.0.0 or some such indication of the *when*?

lib/active_fedora/relation/finder_methods.rb
```diff
@@ -163,6 +168,12 @@ def find_in_batches(conditions, opts = {})
       end
     end
 
+    # @deprecated
+    def find_in_batches(*args)
+      Deprecation.warn(ActiveFedora::Base, '.find_in_batches is deprecated; use .search_in_batches instead')
```
mjgiarlo: Should we add that it'll be going away in 11.0.0 or some such indication of the *when*?

mjgiarlo:
Should we clarify *when* the new deprecations will go into effect? Other than that, I'm :+1: on the change.

mjgiarlo:
:+1: :shipit: 

Thanks, @cbeer

PR #1011, opened 2016-03-15
Extract SolrHit class to wrap Solr response documents

cbeer:


lib/active_fedora/solr_hit.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    attr_reader :document
+
+    def initialize(document)
+      super
+      @document = document
+    end
+
+    def id
+      document[ActiveFedora.id_field]
```
cbeer: I went back and forth about whether this should be `document` or `self`. I landed on `document`, because I anticipate (some day) discouraging consumers from trying to access the raw solr response (and strongly suggesting `instantiate_with_json` or `reify` instead). I addressed all non-test hash access in ActiveFedora, so I'm optimistic that this may be possible.

awead: :+1: I try to avoid doing that as well.

awead:
@cbeer any similarities/alignments between SolrHit and Blacklight::Solr::Response ?

cbeer:
They sure are similar, but without introducing a dependency on blacklight (or extracting all the model stuff  into a separate gem), I'm not sure what to do about it now.

awead:
@cbeer yeah, I wasn't suggesting that we go down that road, yet at least, but just wanted to make sure I understood. I'm :+1: on the PR. @jcoyne ?

PR #1012, opened 2016-03-16
Add default scopes

jcoyne:
These are used by GlobalID, which we should add in the future so we can
use ActiveFedora with ActiveJob


jcoyne:
This is odd:
```
rspec spec/unit/callback_spec.rb:40 spec/unit/scoping_spec.rb
Run options: include {:locations=>{"./spec/unit/callback_spec.rb"=>[27]}}
.F.
```
But no failures if I just call:
```
rspec spec/unit/scoping_spec.rb
..
```

mjgiarlo:
:+1:

PR #1021, opened 2016-03-16
Tell contained resources what container saves them

tpendragon:


mjgiarlo:
:ok_hand:

spec/integration/contains_association_spec.rb
```diff
@@ -16,7 +16,9 @@ class Source < ActiveFedora::Base
       s = Source.new
       s.sub_resource.title = "Test"
       expect(s.sub_resource).not_to be_persisted
+      allow(s.sub_resource).to receive(:save).and_call_original
       expect { s.save }.not_to raise_error
+      expect(s.sub_resource).to have_received(:save).with(container: s)
```
jcoyne: I'm not sure we should be using spys in an integration test. Can you write a model that would actually use the new behavior?

tpendragon: Sure, there ya go.

jcoyne:
:+1:

tpendragon:
This isn't useful to the AF::Aggregations PR anymore, so closing it.

PR #1022, opened 2016-03-16
Remove or deprecate unused code

cbeer:


lib/active_fedora/associations/belongs_to_association.rb
```diff
@@ -36,16 +36,6 @@ def remove_keys
           owner[reflection.foreign_key] = nil
         end
 
-        # Constructs a query that checks solr for the correct id & class_name combination
-        def construct_query(ids)
```
cbeer: This seems vaguely important, but doesn't seem to be used..

jcoyne: It was added here: https://github.com/projecthydra/active_fedora/commit/8ac538eb4442bde23e561735e00d28ff50d55744

jcoyne: It should have been removed here: https://github.com/projecthydra/active_fedora/commit/bbf94f4354b83f0e3463b0ab19e3bb86b63203e8

PR #1024, opened 2016-03-17
Begin updating associations to latest Rails conventions

cbeer:
Slowly pulling our associations code up to date with the latest Rails developments.


jcoyne:
Nice work!

cbeer:
Getting there. There's some nasty work detangling our mangled reflections, but then the first round might be shippable.

cbeer:
I've started running into serious trouble trying to make more progress here; it might make sense to ship this as-is, and iterate on it.

PR #1028, opened 2016-03-18
Update callbacks with upstream improvements

cbeer:


lib/active_fedora/file_persistence.rb
```diff
@@ -0,0 +1,27 @@
...
+        return false if content.nil?
+        ldp_source.content = content
+        ldp_source.create do |req|
+          req.headers.merge!(ldp_headers)
+        end
+        refresh
+      end
+
+      def _update_record(_options = {})
+        return false unless content_changed?
```
cbeer: In 8fd3c606364b550b654b20fa227155ab3cc9ed7a, I've changed this from `true` to `false` (to cancel after_update callbacks, like we're doing above). It seems like this was introduced in c76c6ca37d58d386216f99c7c98a0acf0bf779d0, but it's unclear why create and update are different. @hectorcorrea @jcoyne?

jcoyne: I agree, it should return true if there is nothing to save as ActiveRecord does. See https://gist.github.com/jcoyne/074969310111653ae02c

jcoyne: create record is different because it would not save anything if there is no content to save, that would effectively be an error.

jcoyne: @cbeer I don't really care what `_update_record` returns as long as `save()` returns true in this instance.

cbeer: :ok_hand: I can work with that.

lib/active_fedora/persistence.rb
```diff
@@ -70,6 +70,17 @@ def destroy(*opts)
       delete(*opts)
     end
 
+    # Deletes the record in the database and freezes this instance to reflect
+    # that no changes should be made (since they can't be persisted).
+    #
+    # There's a series of callbacks associated with #destroy!. If the
+    # <tt>before_destroy</tt> callback throws +:abort+ the action is cancelled
+    # and #destroy! raises ActiveRecord::RecordNotDestroyed.
```
jcoyne: Change this line and the next to ActiveFedora instead of ActiveRecord

lib/active_fedora/file_persistence.rb
cbeer: I looked at how Rails handles callbacks -- they don't suppress the `after_update` hook if nothing changed (and, presumably, neither do we, in `AF::Persistence`). I've updated this behavior in 5bae75f.

PR #1033, opened 2016-03-21
Update AF::AttributeMethods with upstream changes

cbeer:


lib/active_fedora/core_core.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    # Returns true if +comparison_object+ is the same exact object, or +comparison_object+
+    # is of the same type and +self+ has an ID and it is equal to +comparison_object.id+.
+    #
+    # Note that new records are different from any other record by definition, unless the
+    # other record is the receiver itself.
+    #
+    # Note also that destroying a record preserves its ID in the model instance, so deleted
+    # models are still comparable.
+    def ==(other)
```
cbeer: Note that the definition of `==` is slightly different than before, but now matches upstream.

lib/active_fedora/core_core.rb
```diff
@@ -0,0 +1,67 @@
+module ActiveFedora
+  module CoreCore
```
jcoyne: Can we call it AF::Common instead? I'm not sure CoreCore is right.

cbeer: :ok_hand: I'm hoping it's only a temporary appearance.

PR #1048, opened 2016-03-28
Align new aggregation associations with existing conventions

cbeer:


lib/active_fedora/reflection.rb
```diff
@@ -477,5 +585,66 @@ def build_index_config
...
+      def collection?
+        false
+      end
+
+      def association_class
+        Associations::SingularRDF
+      end
+    end
+
+    class FilterReflection < ActiveFedora::Reflection::AssociationReflection
```
mjgiarlo: Does this (and line 624) need the explict `ActiveFedora::Reflection` namespacing (e.g., because without it an AR constant is resolved)? I see bare `AssociationReflection`s in other places in this file.

mjgiarlo:
One question. Otherwise :+1: and :clap:!

lib/active_fedora/reflection.rb
cbeer: :+1:

mjgiarlo:
Anything to add, @jcoyne?

PR #1052, opened 2016-04-06
follow redirects for content streaming, closes #1051

### LINKED ISSUE #1051: Streamed content doesn't follow redirects
When a File has a mime-type of 'message/external-body; access-type=URL; url="..."' a HTTP request will result in a resonse status of HTTPRedirection. The File::Streaming module however only checks for a success response and raises "Couldn't get data from Fedora..." in any other case.

When a HTTPRedirection response is received, it should make a new request for the data at the response's location (where the redirect points) and stream that data back instead.

lbiedinger:


lib/active_fedora/file/streaming.rb
```diff
@@ -22,13 +22,24 @@ def initialize(uri, headers)
...
-            yield chunk
+          case response
+          when Net::HTTPSuccess
+            response.read_body do |chunk|
+              yield chunk
+            end
+          when Net::HTTPRedirection
+            no_of_requests_limit -= 1
+            @uri = URI(response["location"])
+            each no_of_requests_limit do |chunk|
```
jcoyne: If you change the signature of `each` to take a `&block`, you can just pass the block back to `each` like this: `each(no_of_requests_limit, &block)`

lib/active_fedora/file/streaming.rb
```diff
@@ -22,13 +22,24 @@ def initialize(uri, headers)
...
-          response.read_body do |chunk|
-            yield chunk
+          case response
+          when Net::HTTPSuccess
+            response.read_body do |chunk|
+              yield chunk
+            end
+          when Net::HTTPRedirection
+            no_of_requests_limit -= 1
+            @uri = URI(response["location"])
```
jcoyne: I'm concerned about the unintended consequences of setting `@uri`.  What happens if you call `update(label: 'test')` on this object after streaming?

lbiedinger: the object which has @uri set is the FileBody, which doesn't have an update method. The File's uri isn't changed. Is there anywhere other than the Streaming module where FileBody is used?

jcoyne: :ok_hand:

PR #1053, opened 2016-04-07
Fedora 4.5.1-RC compatibility

escowles:
* Updates test to not rely on presence of mix:versionable triple (being removed in Fedora 4.5.1)

### DIFF
--- a/spec/integration/versionable_spec.rb
+++ b/spec/integration/versionable_spec.rb
@@ -179,8 +179,8 @@ describe ActiveFedora::Versionable do
           @original_size = subject.size
         end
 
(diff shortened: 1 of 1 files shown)


escowles:
When run against Fedora 4.5.1-RC, I see three test failures: https://gist.github.com/escowles/a2d88856fa8b9cefd5f84dd42ae15cc4

These failures all stem from checking for a mix:versionable triple, which is being removed in 4.5.1.  I've updated the tests to check for the fcr:hasVersions triple instead.

jcoyne:
Thanks for this work @escowles!

PR #1056, opened 2016-04-11
SolrHit#instantiate_with_json should create persisted objects

### LINKED ISSUE #1054: model returned by ActiveFedora::SolrHit. instantiate_with_json is persisted? -> false
active-fedora (9.10.4)
ldp (0.5.0)
rails (4.2.5.1)

this breaks polymorphic path builders, which will not create a show path for an unpersisted model.

cbeer:
Fixes #1054

### DIFF
--- a/lib/active_fedora/solr_hit.rb
+++ b/lib/active_fedora/solr_hit.rb
@@ -21,6 +21,7 @@ module ActiveFedora
     attr_reader :document
 
     def initialize(document)
+      document = document.with_indifferent_access
       super
       @document = document
     end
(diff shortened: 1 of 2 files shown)


cbeer:
@barmintor

jcoyne:
What? That was the bug?  :+1:

cbeer:
Not sure. It was one thing I tripped across trying to recreate the bug in the first place, and I'm hoping it's related.

barmintor:
I am chasing down various dependency conflicts preventing me from testing this in my sufia7 app, the latest is that hydra-access-controls has not been released and is not a github repo, so it's released pin of AF ~> 9.10.x is in effect, barring AF master.

PR #1063, opened 2016-04-19
Update graph instead of destroying it

narogers:
By not destroying the entire graph every time performance grows linearly instead of in an exponential fashion. It still needs to be tested on a large scale ( > 1000 members) but for hasCollectionMember relationships of < 500 it goes from 1200 seconds to 192 in my test environment.


jcoyne:
Thanks @narogers! This looks great.

PR #1072, opened 2016-05-05
Renamed `contains` to `has_subresource`

jcoyne:
`contains` is deprectated until AF 10. The plan is to add a true basic
containment, where I post to the BasicContainer to get a URI (rather
than knowing the relative path ahead of time).

### DIFF
--- a/lib/active_fedora/associations.rb
+++ b/lib/active_fedora/associations.rb
@@ -24,7 +24,7 @@ module ActiveFedora
     autoload :HasManyAssociation
     autoload :BelongsToAssociation
     autoload :HasAndBelongsToManyAssociation
-    autoload :BasicContainsAssociation
+    autoload :HasSubresourceAssociation
     autoload :DirectlyContainsAssociation
     autoload :DirectlyContainsOneAssociation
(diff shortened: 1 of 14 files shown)


lib/active_fedora/associations.rb
```diff
@@ -45,7 +45,7 @@ module Builder
       autoload :BelongsTo,           'active_fedora/associations/builder/belongs_to'
       autoload :HasMany,             'active_fedora/associations/builder/has_many'
       autoload :HasAndBelongsToMany, 'active_fedora/associations/builder/has_and_belongs_to_many'
-      autoload :Contains,            'active_fedora/associations/builder/contains'
+      autoload :BasicContainsOne,    'active_fedora/associations/builder/basic_contains_one'
```
cbeer: I'm confused by the name `BasicContainsOne`. This seems like it holds a sub-resource at a known location (which we should feel bad about doing, instead of defining and using predicates..); maybe `has_subresource` would be a better name?

jcoyne: 👌

PR #1076, opened 2016-05-10
Remove deprecations in preparation for AF 10.0

jcoyne:


lib/active_fedora/associations/builder/association.rb
```diff
@@ -67,18 +63,6 @@ def self.better_name(name)
...
     end
 
-    def self.translate_property_to_predicate(options)
-      return unless options[:property]
-      Deprecation.warn Association, "the :property option to `#{model}.#{macro} :#{name}' is deprecated and will be removed in active-fedora 10.0. Use :predicate instead", caller(5)
-      options[:predicate] = predicate(options.delete(:property))
-    end
-
-    # Returns the RDF predicate as defined by the :property attribute
-    def self.predicate(property)
```
cam156: @jcoyne This method does not have a deprecation warning.  Did you intend to remove it?

lib/active_fedora/attributes.rb
```diff
@@ -49,21 +45,10 @@ def []=(key, value)
...
       end
     end
 
-    # @return [Boolean] true if there is an reader method and it returns a
-    # value different from the new_value.
-    def value_has_changed?(field, new_value)
-      new_value != array_reader(field)
-    end
-
-    def mark_as_changed(field)
```
cam156: @jcoyne No deprecation warnings for this method or value_has_changed.  Was this intentional?

lib/active_fedora/file.rb
```diff
@@ -34,7 +33,7 @@ class File
     # @param [Hash] _options
     # @yield [self] Yields self
     # @yieldparam [File] self the newly created file
-    def initialize(parent_or_url_or_hash = nil, path = nil, _options = {}, &_block)
+    def initialize(parent_or_url_or_hash = nil, _path = nil, _options = {}, &_block)
```
cam156: @jcoyne would now be the correct time to remove the path parameter since it is not used?

lib/active_fedora/associations/builder/association.rb
jcoyne: Yes, it's only used by a deprecated method.  We didn't add a deprecation to that method too, otherwise the user would get two deprecation warnings when they did the deprecated behavior.

lib/active_fedora/file.rb
jcoyne: Prior to releasing 10.0 is a good time, but not necessarily in this PR.

lib/active_fedora/attributes.rb
jcoyne: This was only used by classes that were deprecated and removed, so it's no longer necessary.

jcoyne:
@cam156 okay, I couldn't wait. I refactored `File#new` to not have the unused params

lib/active_fedora/associations/builder/association.rb
```diff
@@ -10,7 +10,7 @@ class << self
     # configure_dependency
     def self.build(model, name, options, &block)
       if model.dangerous_attribute_method?(name)
-        Deprecation.warn(ActiveFedora::Base, "You tried to define an association named #{name} on the model #{model.name}, but " \
+        ActiveFedora::Base.logger.error("You tried to define an association named #{name} on the model #{model.name}, but " \
```
cbeer: Is logging enough to expose this to an implementor, or is it just going to get lost in the rails log?

jcoyne: I'm not sure that a deprecation is better. Deprecation seems incorrect as this is not a deprecation. Are you suggesting that we raise an error?

cam156:
👍 from me!

lib/active_fedora/associations/builder/association.rb
cbeer: Let me check, but I think that's the upstream behavior.

cbeer: Yes, Rails raises an exception:

https://github.com/rails/rails/blob/master/activerecord/lib/active_record/associations/builder/association.rb#L22

We must have downgraded it to a deprecation (and I should have noted that in a comment..)

jcoyne: Yeah they raise: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/associations/builder/association.rb#L23

awead:
Lots of removed code + green build = 👍

lib/active_fedora/associations/builder/association.rb
jcoyne: I'll do this as a separate PR

PR #1077, opened 2016-05-10
Metadata class factory

awead:
Allows you to add additional schemas of property definitions on your non-RDF sources such as ActiveFedora::File.


jcoyne:
Is this still WIP?

awead:
@jcoyne It's probably ok to go. I've tested against hydra-pcdm and it looks good. I'm in the middle of testing it against hydra-works and it seems to be good so far.

PR #1081, opened 2016-05-17
Setting the download directory so you only get one instance downloaded…

### LINKED ISSUE #1080: setting instance_dir causes solr to download for each project
https://github.com/projecthydra/active_fedora/blob/master/lib/generators/active_fedora/config/solr/templates/solr_wrapper_test.yml#L4

@mark-dce @cam156 was this your intent?

cam156:
per machine. fixes #1080

### DIFF
--- a/lib/generators/active_fedora/config/fedora/templates/.fcrepo_wrapper
+++ b/lib/generators/active_fedora/config/fedora/templates/.fcrepo_wrapper
@@ -2,3 +2,4 @@
 port: 8984
 enable_jms: false
 fcrepo_home_dir: tmp/fcrepo4-development-data
+download_dir: /tmp
--- a/lib/generators/active_fedora/config/fedora/templates/fcrepo_wrapper_test.yml
+++ b/lib/generators/active_fedora/config/fedora/templates/fcrepo_wrapper_test.yml
@@ -2,3 +2,4 @@
 port: 8986
 enable_jms: false
 fcrepo_home_dir: tmp/fcrepo4-test-data
+download_dir: /tmp
--- a/lib/generators/active_fedora/config/solr/templates/.solr_wrapper
+++ b/lib/generators/active_fedora/config/solr/templates/.solr_wrapper
@@ -2,6 +2,7 @@
 # version: 6.0.0
 # port: 8983
 instance_dir: tmp/solr-development
+download_dir: /tmp
 collection:
   persist: true
   dir: solr/config/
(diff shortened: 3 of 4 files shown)


lib/generators/active_fedora/config/fedora/templates/.fcrepo_wrapper
```diff
@@ -2,3 +2,4 @@
 port: 8984
 enable_jms: false
 fcrepo_home_dir: tmp/fcrepo4-development-data
+download_dir: /tmp
```
jcoyne: I wonder if these should be `<%= ENV['TMPDIR'] %>`? @cbeer ?

cbeer: That'd certainly be better (and cross platform..)... although isn't this the same as the default behavior upstream?

mjgiarlo: `<%= ENV.fetch('TMPDIR', '/tmp') %>` ?

jcoyne: @cbeer Yes, that is curious.  Why is it downloading then?

jcoyne:
@cam156 are you sure this fixes the issue?

Prior to using this change I see:

```
require 'solr_wrapper'
=> true
irb(main):002:0> SolrWrapper.instance(config: 'config/solr_wrapper_test.yml').config.instance_dir
=> "tmp/solr-test"
irb(main):003:0> SolrWrapper.instance(config: 'config/solr_wrapper_test.yml').config.download_dir
=> "/var/folders/9t/rygbnddx0b1ckw6tjs3m18qm0000gq/T"
```

So it appears as though it's already looking in the tmp directory for the download_dir.

cam156:
@jcoyne looking at the default here: https://github.com/cbeer/solr_wrapper/blob/master/lib/solr_wrapper/configuration.rb#L58-L60
If you are in a rails app it uses tmp instead of .tmp
So depending on where you run the wrapper it behaves differently.  So I am not 100% sure, but I would believe it to work...

jcoyne:
@cam156 Yes, I see that. I wonder if it makes more sense to just remove that logic from solr_wrapper?

cam156:
@jcoyne I would be ok with that approach.

mjgiarlo:
@cam156 @jcoyne so am I hearing that this will be fixed in solr_wrapper, and no related change is needed in AF? or?

jcoyne:
I think this should get fixed in solr_wrapper. It's a broader issue than just AF.

mjgiarlo:
OK. @cam156 should we close this issue?

cam156:
@mjgiarlo Is there an issue on Solr wrapper?  I would leave it open until there is...

mjgiarlo:
@cam156 There are 0 open issues on solr_wrapper, so no.

PR #1091, opened 2016-05-31
AF::File#save! should persist metadata

jcoyne:
Previously metadata was only persisted when `save()` was called,
`save!()` was not persisting.

### DIFF
--- a/lib/active_fedora/file.rb
+++ b/lib/active_fedora/file.rb
@@ -48,12 +48,6 @@ module ActiveFedora
       end
     end
 
-    def save
-      super.tap do
-        metadata.save if metadata.changed?
-      end
-    end
-
     def described_by
       raise "#{self} isn't persisted yet" if new_record?
       links['describedby'].first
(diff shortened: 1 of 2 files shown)


lib/active_fedora/file.rb
```diff
@@ -176,6 +170,12 @@ def prefix(path)
 
     private
 
+      def create_or_update(*options)
+        super.tap do
+          metadata.save if metadata.changed?
```
cbeer: (Without looking), should this be `#save!` and/or do we need some error handling logic if `#save` fails?

jcoyne: It's possible, but that should be in a different PR.  This PR is about moving the logic to be called when `save` or `save!` are called.

awead:
👍 if you can link a new issue for the `save!` question.

lib/active_fedora/file.rb
jcoyne: `ActiveFedora::WithMetadata::MetadataNode` does not respond to `save!` and does not have any error handling.

jcoyne:
@awead I don't really even know where to begin on this ticket: https://github.com/projecthydra/active_fedora/issues/1092

awead:
@jcoyne yeah, it's possible we might just close it, but the ticket would have the discussion. I'll merge this PR now since we've marked the save! issue for later resolution, even if the resolution is we don't want to do anything.

lib/active_fedora/file.rb
cbeer: 👌

PR #1098, opened 2016-06-16
Deprecate datastreams

jcoyne:
Datastreams no longer need to be part of ActiveFedora. They are just ActiveFedora::Files.

### DIFF
--- a/lib/active_fedora/nom_datastream.rb
+++ b/lib/active_fedora/nom_datastream.rb
@@ -3,10 +3,16 @@ require "nom"
 module ActiveFedora
   class NomDatastream < File
     include Datastreams::NokogiriDatastreams
+    extend Deprecation
 
     def self.set_terminology(options = {}, &block)
       @terminology_options = options || {}
       @terminology = block
+    end
+
+    def initialize(*args)
+      super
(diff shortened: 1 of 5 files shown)


mjgiarlo:
Folks are going to ask how they can continue to use OM or something like it to create terminologies for their XML-holding `AF::Files`. Will this behavior change when OM datastreams are deprecated?

jcoyne:
In a future version you can use https://github.com/projecthydra-labs/active_fedora-datastreams

barmintor:
@mjgiarlo we need an issue that blocks AF 11 release to document metadata default and index strategy definitions for AF::File.

lib/active_fedora/nom_datastream.rb
```diff
@@ -3,12 +3,18 @@
...
+    extend Deprecation
 
     def self.set_terminology(options = {}, &block)
       @terminology_options = options || {}
       @terminology = block
     end
 
+    def initialize(*args)
+      super
+      Deprecation.warn(NomDatastream, "NomDatastream is deprecated and will be removed in ActiveFedora 11")
```
atz: 👍

PR #1101, opened 2016-06-23
Jruby travis

atz:
Old jruby branch was old.  Can't do just `jruby` because that gets **1.x** version of Ruby. We require **2.x**.   

Travis has jruby build issues w/ RVM images not including bundler (and Travis failing to backfill where bundler is expected).  Workaround allows app testing to happen, but we get:

```
1065 examples, 475 failures, 1 pending
```

Unclear how much is due to CI environment vs. actual jruby compatibility issues.  But there's a lot to look at.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ cache: bundler
 sudo: false
 rvm:
   - 2.3.0
+  - jruby-head
 
 notifications:
   irc: "irc.freenode.org#projecthydra"
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ cache: bundler
 sudo: false
 rvm:
   - 2.3.0
-  - jruby-head
+  - jruby-9.0.5.0
 
 notifications:
   irc: "irc.freenode.org#projecthydra"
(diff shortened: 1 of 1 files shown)


mjgiarlo:
I wonder how committed we ought to be to supporting JRuby.

no-reply:
You might want to add `- jruby-9.0.0.0` in addition to `- jruby-head`. On other projects (e.g. [Hamster](https://travis-ci.org/hamstergem/hamster/builds/137495918)) I've found that `jruby-head` fails in a bunch of cases where released versions do not.

I'd suggest giving this a go with allowed failures on `head`, but testing some significant version genuinely.

atz:
@no-reply: yeah, agree re: version.

But [this](https://docs.travis-ci.com/user/languages/ruby#JRuby%3A-C-extensions-support-is-disabled) is probably more significant of a deal for our overall support:

`C extensions are disabled for JRuby on Travis CI`

So we would need to make a gemfile for jruby that did not use C extensions.

atz:
With `jruby-9.0.5.0`, we still get:
```
1065 examples, 476 failures, 1 pending
```
Which is to say, it is 1 degree worse.  The log exceeds 4MB and Travis truncates it.

atz:
@no-reply Seems like we should just declare `jruby` unsupported (by policy or at least _de facto_) and close this.

no-reply:
Closing, per @atz's comment.

I think if we're going to support JRuby formally, we need a champion to (at least) maintain the CI build.

PR #1102, opened 2016-06-28
Fixing rubocop for latest version

cam156:
Fixing rubocop so that other commits to this branch will get a green build.

### DIFF
--- a/spec/integration/scoped_query_spec.rb
+++ b/spec/integration/scoped_query_spec.rb
@@ -39,7 +39,7 @@ describe ActiveFedora::Querying do
         result = ModelIntegrationSpec::Basic.all.to_a
         expect(result).to be_instance_of(Array)
         # this test is meaningless if the array length is zero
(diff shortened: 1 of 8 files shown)


spec/integration/scoped_query_spec.rb
```diff
@@ -39,7 +39,7 @@ def to_solr(doc = {})
         result = ModelIntegrationSpec::Basic.all.to_a
         expect(result).to be_instance_of(Array)
         # this test is meaningless if the array length is zero
-        expect(result.length > 0).to be true
+        expect(!result.empty?).to be true
```
cbeer: `expect(result).not_to be_empty`?

cam156:
@jcoyne created this branch so I can do a small tweak (allow where to join with OR) to active fedora which will allow my export code to be cleaner.  I have another PR ready once this gets merged.  I am also planning to add the tweak to master as I think it would be a good tweak to have into the future.

PR #1103, opened 2016-06-29
Adding the ability to join with OR from a where clause

cam156:
This allows for finding all the objects that are 'this OR that' instead of just 'this AND that'
Added parens around cluases since 'abc' AND  'ddd' OR 'fff' is not the same as 'abc' AND ('ddd' OR 'fff').  This was not an issue before since all the clauses were joined with AND

### DIFF
--- a/lib/active_fedora/scoping/named.rb
+++ b/lib/active_fedora/scoping/named.rb
@@ -13,6 +13,7 @@ module ActiveFedora
         #
         #   fruits = Fruit.all
         #   fruits = fruits.where(color: 'red') if options[:red_only]
+        #   fruits = fruits.where(size: ['big', 'little']) # gets big or little fruit
         #   fruits = fruits.limit(10) if limited?
         #
         # You can define a scope that applies to all finders using
--- a/lib/active_fedora/solr_query_builder.rb
+++ b/lib/active_fedora/solr_query_builder.rb
@@ -32,7 +32,7 @@ module ActiveFedora
       #   # => _query_:"{!raw f=has_model_ssim}info:fedora/afmodel:ComplexCollection" OR _query_:"{!raw f=has_model_ssim}info:fedora/afmodel:ActiveFedora_Base"
       #
       #   construct_query_for_rel [[Book.reflect_on_association(:library), "foo/bar/baz"]]
-      def construct_query_for_rel(field_pairs, join_with = ' AND ')
+      def construct_query_for_rel(field_pairs, join_with = default_join_with)
         field_pairs = field_pairs.to_a if field_pairs.is_a? Hash
         construct_query(property_values_to_solr(field_pairs), join_with)
       end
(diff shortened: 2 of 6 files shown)


cam156:
I will also be creating a PR for master once all the kinks are worked out on this one.

jcoyne:
If you set the default join with to be ' OR ' would that accidentally give you a bigger scope than you intended?  Consider you have a `belongs_to :collection` and `has_many :members`.  
When you build the query for "show me the things that belong to collection X and have a title like `foo`" like this:  `colleciton.members.where(title: 'foo')`  would that still work?

lib/active_fedora/solr_query_builder.rb
```diff
@@ -44,8 +44,14 @@ def construct_query_for_rel(field_pairs, join_with = ' AND ')
       # @example
       #   construct_query([['library_id_ssim', '123'], ['owner_ssim', 'Fred']])
       #   # => "_query_:\"{!raw f=library_id_ssim}123\" AND _query_:\"{!raw f=owner_ssim}Fred\""
-      def construct_query(field_pairs, join_with = ' AND ')
-        pairs_to_clauses(field_pairs).join(join_with)
+      def construct_query(field_pairs, join_with = default_join_with)
+        clauses = pairs_to_clauses(field_pairs)
+        return clauses.join("") if clauses.count <= 1
```
jcoyne: would `clauses.first` be more efficient? Then we don't have to allocate a new string.

lib/active_fedora/scoping/named.rb
```diff
@@ -13,6 +13,7 @@ module ClassMethods
         #
         #   fruits = Fruit.all
         #   fruits = fruits.where(color: 'red') if options[:red_only]
+        #   fruits = fruits.where(size: ['big', 'little'], join_with: ' OR ') #gets big or little fruit
```
jcoyne: In activerecord, passing an array, always produces an `IN` query. This is equivalent to an `OR` in solr.  I think we should do away with the `join_with` argument and say, if you pass in an array, you want an or.  Here's how AR works:

```ruby
User.where(email: ['foo', 'bar'])
  SELECT "users".* FROM "users" WHERE "users"."email" IN ('foo', 'bar')
```

cam156:
@jcoyne The default join_with is still AND.  We are only using OR when the user passes it with a join_with option.
I do think the default OR might give us more than we expect, unless it was only used in the array case.  I am not familiar with all the places this code is used, so I thought it safer to leave the default the same.

cam156:
@jcoyne I changed the default to ' OR ' and there are a bunch of test failures. I think leaving the default as 'AND' in 9.7 is the right move.  We could consider changing the default in master.

jcoyne:
@cam156 right, but that looks like a bad expectation to me. I'd like to see the use case that emerged from.

cam156:
@jcoyne I agree the OR makes more sense, but I am trying to make the smallest change here in the 9.7.x branch so we can reasonably stay in the 9.7.x numbering.  I can test changing to OR as the default with the sufia test, but I'm worried it will have unforeseen consequences in a patch release.

awead:
The current expected behavior is AND. I think we should preserve the expected behavior. With this PR, you now have the option of using OR so I think this PR solves the problem as described.

lib/active_fedora/scoping/named.rb
mjgiarlo: @cam156 what do you say about this?

cam156: @mjgiarlo & @jcoyne: see all the comments below....  I do not think that changing to a OR in this branch is a great idea.

jcoyne:
To me it is clear that our existing API is not behaving correctly, and we should fix it. It is a bug.  We should not introduce a new API which we must support in perpetuity to do exactly how it should have worked in the first place.

jcoyne:
When I added this test https://github.com/psu-stewardship/active_fedora/commit/d0397ab800d90e119a9647e8f47b1f19f3b76ff1#diff-d6939355441580118b5afdbbd0a98049R44 , which will fail if we make my suggested change, it was not because I was being intentional about supporting that feature, it was more of just testing the existing behavior.

cam156:
@jcoyne 👌 If this is a bug fix then I am good with it.  I will make the change.  We can always bug fix again.  So you would like to drop the optional join_with and just change the where clause to use OR?

cam156:
@jcoyne Better?

PR #1104, opened 2016-06-29
Update Active Triples

tpendragon:
Few compatibility issues here that would have to be removed to support
this:

1. Order of statements that come out of property accessors is now truly
random, because of RDF.rb 2.0. A LOT of tests will have to change, and
some indexing may be unexpected.

2. There's no longer support in AT for one predicate having multiple
accessors. The logic behind that work in the past was shifty at best,
and often would result in lost data. It's for the best that it's gone -
but it's important to note.

3. This is dependent on a branch in AT that isn't merged and is in discussion, but I thought it'd be good to have this PR up.

### DIFF
--- a/active-fedora.gemspec
+++ b/active-fedora.gemspec
@@ -20,8 +20,8 @@ Gem::Specification.new do |s|
   s.add_dependency 'nom-xml', '>= 0.5.1'
   s.add_dependency "activesupport", '>= 4.2.4', '< 6'
   s.add_dependency "activemodel", '>= 4.2', '< 6'
-  s.add_dependency "active-triples", '~> 0.7.1'
-  s.add_dependency "rdf-rdfxml", '~> 1.1'
+  s.add_dependency "active-triples", '~> 0.10.0'
+  s.add_dependency "rdf-rdfxml"
   s.add_dependency "linkeddata"
   s.add_dependency "deprecation"
   s.add_dependency "ldp", '~> 0.5.0'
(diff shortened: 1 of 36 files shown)


awead:
@tpendragon should this be considered ActiveFedora 11?

tpendragon:
@awead Yes, this'll require a major version increment.

tpendragon:
I think there's a bug here that's showing up in Hydra Works that I'm working through now.

awead:
I'm 👍 on principle. The reality is that this will force an issue with our current userbase because they've come to expect a preserved order among multiple terms. Because that's currently even a false expectation that's just being hidden somewhat, I believe it's better to force the issue sooner than later, but we need a plan to address their concerns.

tpendragon:
> we need a plan to address their concerns.

I think we'll have to address this in things like Curation Concerns via sorting in the solr document (or presenters)

no-reply:
> Because that's currently even a false expectation that's just being hidden somewhat, I believe it's better to force the issue sooner than later, but we need a plan to address their concerns.

This does seem like it calls for some thoughtful communication. I've added it to the [Hydra Tech agenda for 7/6](https://wiki.duraspace.org/display/hydra/Hydra+Tech+Call+2016-07-06). I'm happy to lead any discussion on this. We can follow-up with an email to various lists, as appropriate after the call.

I've made some of the key points about why it's probably best to bite the bullet up-front in the agenda, so folks will have some time to think on it.

tpendragon:
@projecthydra/sufia-code-reviewers This PR is ready. I think merging this (and then merging https://github.com/projecthydra/active_fedora/tree/test_speed_updates) represents ActiveFedora 11.

The ordering issues are going to take some time to work through. However, ingests with this and the speed improvements are at least twice as fast in Plum - getting to the most recent version of our dependencies and getting that speed boost seems worth it to me.

awead:
👍 but I'm not going to 🔫

PR #1107, opened 2016-07-08
Update Active Triples

jcoyne:
Few compatibility issues here that would have to be removed to support
this:

1. Order of statements that come out of property accessors is now truly
random, because of RDF.rb 2.0. A LOT of tests will have to change, and
some indexing may be unexpected.

2. There's no longer support in AT for one predicate having multiple
accessors. The logic behind that work in the past was shifty at best,
and often would result in lost data. It's for the best that it's gone -
but it's important to note.


jcoyne:
Can we do #1098 first, then release 10.1.0, then we'll merge this and master will be  11.0.0?

tpendragon:
@jcoyne @awead So, can we merge this now that #1098 is merged?

PR #1109, opened 2016-07-11
Performance Improvements

tpendragon:
Shares an in-memory repository between the LDP graph and the AF::Base's
resource. Reduces graph copying by a significant amount.

Also adds an ids reader to ordered associations which doesn't instantiate objects.

### DIFF
--- a/lib/active_fedora/fedora_attributes.rb
+++ b/lib/active_fedora/fedora_attributes.rb
@@ -43,7 +43,7 @@ module ActiveFedora
       # Appending the graph at the end is necessary because adding it as the
       # parent leaves behind triples not related to the ldp_source's rdf
       # subject.
-      @resource ||= self.class.resource_class.new(@ldp_source.graph.rdf_subject, @ldp_source.graph) << @ldp_source.graph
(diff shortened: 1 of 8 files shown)


lib/active_fedora/fedora_attributes.rb
```diff
@@ -43,7 +43,7 @@ def resource
       # Appending the graph at the end is necessary because adding it as the
       # parent leaves behind triples not related to the ldp_source's rdf
       # subject.
-      @resource ||= self.class.resource_class.new(@ldp_source.graph.rdf_subject, @ldp_source.graph) << @ldp_source.graph
+      @resource ||= self.class.resource_class.new(@ldp_source.graph.rdf_subject, data: @ldp_source.graph.send(:graph).data)
```
jcoyne: Should a ticket be opened upstream to change the visibility to public?

tpendragon: @no-reply Should we make AT::Resource's graph public, or just delegate data?

tpendragon: https://github.com/ActiveTriples/ActiveTriples/issues/222

PR #1119, opened 2016-08-02
Fixing for latest rubocop-rspec

cam156:
Working on fixing the test for the latest rubocop-rspec errors

### DIFF
--- a/spec/integration/associations_spec.rb
+++ b/spec/integration/associations_spec.rb
@@ -17,7 +17,7 @@ describe ActiveFedora::Base do
     end
 
     let(:person) { Person.create }
-    let(:book) { Book.new(author: person) }
+    subject(:book) { Book.new(author: person) }
 
     it "goes" do
       book.save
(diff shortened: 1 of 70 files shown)


cam156:
I'm putting this out so that everyone knows I'm working on the issue, but I have not completely fixed all the instances....

spec/integration/associations_spec.rb
```diff
@@ -524,18 +524,19 @@ class Page < ActiveFedora::Base
...
       describe "removing association" do
-        subject { LibraryBook.create }
-        before do
-          @p1 = Page.create
-          @p2 = Page.create
-          subject.pages << @p1 << @p2
-          subject.save!
+        let(:p1) { Page.create }
+        let(:p2) { Page.create }
+        let(:book) do
```
backus: FYI part of the idea of the `NamedSubject` cop was that you would change `subject { LibraryBook.create }` to `subject(:book) { LibraryBook.create }`. This is fine too but `subject(:name)` might be nice

cam156: @backus Thanks for the tip!

cam156:
@jcoyne, @backus I think this is ready.

PR #1120, opened 2016-08-03
Adding in the terms handler so that generated apps inculde this by default

cam156:
This PR is dependant on #1119 

Also adding in a few additional changes from Sufia that seemed reasonable

This change should fix the issue that some people have been seeing when running tests in other gems:

```
   Failure/Error:
       json = solr_connection.get 'terms', params: { 'terms.fl' => term,
```

The solr config that is currently generated by active fedora does not include the terms configuration.

### DIFF
--- a/lib/generators/active_fedora/config/solr/templates/solr/config/solrconfig.xml
+++ b/lib/generators/active_fedora/config/solr/templates/solr/config/solrconfig.xml
@@ -187,6 +187,17 @@
     </lst>
   </requestHandler>
 
+  <searchComponent name="termsComponent" class="solr.TermsComponent" />
+
+  <requestHandler name="/terms" class="solr.SearchHandler">
+    <lst name="defaults">
+      <bool name="terms">true</bool>
+    </lst>
+    <arr name="components">
+      <str>termsComponent</str>
+    </arr>
+  </requestHandler>
+
 <!-- Spell Check
 
         The spell check component can return a list of alternative spelling


lib/generators/active_fedora/config/solr/templates/solr/config/solrconfig.xml
```diff
@@ -102,6 +102,7 @@
        -->
         <str name="qf">
           id
+          active_fedora_model_ssi
```
jcoyne: what does this have to do with the terms handler?

jcoyne:
This should be rebased now.

lib/generators/active_fedora/config/solr/templates/solr/config/solrconfig.xml
cam156: @jcoyne nothing as far as I could tell, but it seemed reasonable...

cam156:
@jcoyne rebased.  I got the additional changes, like active_fedora_model_ssi, from the Sufia configuration.  I am willing to remove them if you would like.

jcoyne:
@cam156 Yes, I think the PR should be as narrowly scoped as possible.

jcoyne:
If people have a poorly configured solr aren't they still going to have trouble when they try to do full text extraction?  That's a solr config thing that lives only in CurationConcerns/Sufia, right?   I wonder if we shouldn't make it easer for the correct configs to be created from Sufia, rather than make changes here.

lib/generators/active_fedora/config/solr/templates/solr/config/solrconfig.xml
mjgiarlo: I wonder if this is cruft. `active_fedora_model_ssi` does not seem to be widely used; I could only find two references to that field in Sufia's dependencies:

 * https://github.com/projecthydra/curation_concerns/blob/master/spec/views/curation_concerns/base/_member.html.erb_spec.rb#L6
 * https://github.com/projecthydra/hydra-head/blob/master/hydra-core/spec/helpers/blacklight_helper_spec.rb#L23-L25

cam156:
@jcoyne I thought about changing the configs just in Sufia.  It seemed like this issue was giving multiple people at multiple layers heartburn, I put in a root fix.
@jcoyne I'll remove the extras...

mjgiarlo:
@cam156 :heart: :fire: --

cam156:
@jcoyne or others... The extras have been removed.  Merge or Move?

awead:
I vote merge.

jcoyne:
What about the extraction handler?

awead:
Meh. I can go either way, but the ticket is for terms handler.

jcoyne:
Yeah, but we should either have a configuration that has all the needs of Sufia, or don't bother, right? Why have a configuration that only has some of the needs met?

awead:
OK, that makes sense. The extraction is already a part of Solr, it's just a matter of turning it on, right? I just want to make sure that we're not requiring additional things to be downloaded/installed that ActiveFedora isn't using directly.

jcoyne:
AFAIK, everything is already downloaded, we just need the correct config. Up until now I believe that's lived in CC/Sufia.

mjgiarlo:
I support this PR, and wonder if it should also include the same change to `solr/config/solrconfig.xml` (or wipe out `solr/config/` entirely -- see #1125).

PR #1124, opened 2016-08-10
Align solr config from Sufia with what's in AF

mjgiarlo:


mjgiarlo:
Maybe #1120 already takes care of this, or should the two solrconfig.xml files be identical? (I'd also like to make the #1120 change in `10.1-stable`.)

jcoyne:
Shouldn't this PR be against the 10.2-stable branch?  10.1.0 has already been released

mjgiarlo:
Yep, it should, @jcoyne

PR #1130, opened 2016-08-10
Remove Datastreams

jcoyne:
They have been moved to
https://github.com/projecthydra-labs/active_fedora-datastreams

Deprecations went in with #1098


no-reply:
I'm :+1: on this. There was a -0 in the Hydra-Tech thread on the topic, so I'm inclined to look for another :+1: to get this in.

Any takers?

PR #1131, opened 2016-08-10
Remove linkeddata.

jcoyne:
This is a meta gem. We don't want to foist all these dependencies on a user

This should eliminate problems like:
```
Your bundle requires gems that depend on each other, creating an
infinite loop.

Please remove either gem 'linkeddata' or gem 'rdf-vocab' and try again.
```


mjgiarlo:
See also #1132

PR #1135, opened 2016-08-11
solr_wrapper now uses the generated solr configs for testing AF instead of a separate directory tree of Solr configs

### LINKED ISSUE #1125: Question: do we need to store identical Solr configs in two places?
Those being `lib/generators/active_fedora/config/solr/templates/solr/config/` and `solr/config/`

mjgiarlo:
It also provides a .solr_wrapper config that points solr_wrapper at the right directory. Fixes #1125


README.md
```diff
@@ -75,7 +75,7 @@ rake active_fedora:ci
 If you want to run the tests manually, follow these instructions:
 
 ```bash
-solr_wrapper -d solr/config/ --collection_name hydra-test
+solr_wrapper
 ```
 
 You can switch `hydra-test` to `hydra-dev` if you want to create a collection for development.
```
atz: Nothing to switch now, should update this sentence also.

PR #1141, opened 2016-08-18
Lookup inverse using predicate

jcoyne:
Prevously we were using property, but we recently stopped using
property, so it isn't finding the correct relation.


jcoyne:
I suspect we need to backport this to the 10.x branch too.

no-reply:
:+1:

PR #1147, opened 2016-09-08
Querying for relationships should use solr raw query

jcoyne:
Rather than a fielded query

### DIFF
--- a/lib/active_fedora/solr_query_builder.rb
+++ b/lib/active_fedora/solr_query_builder.rb
@@ -11,28 +11,29 @@ module ActiveFedora
         "{!terms f=#{ActiveFedora.id_field}}#{ids.join(',')}"
       end
 
-      # Create a query with a clause for each key, value
+      # Create a raw query with a clause for each key, value
       # @param [Hash, Array<Array<String>>] field_pairs key is the predicate, value is the target_uri
       # @param [String] join_with ('AND') the value we're joining the clauses with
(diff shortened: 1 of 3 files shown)


lib/active_fedora/solr_query_builder.rb
```diff
@@ -11,28 +11,29 @@ def construct_query_for_ids(id_array)
...
       # @param [Array<Array>] field_pairs a list of pairs of property name and values
       # @param [String] join_with ('AND') the value we're joining the clauses with
+      # @param [String] type ('field') The type of query to run. Either 'term' or 'field'
       # @return [String] a solr query
       # @example
       #   construct_query([['library_id_ssim', '123'], ['owner_ssim', 'Fred']])
       #   # => "_query_:\"{!field f=library_id_ssim}123\" AND _query_:\"{!field f=owner_ssim}Fred\""
-      def construct_query(field_pairs, join_with = default_join_with)
-        clauses = pairs_to_clauses(field_pairs)
+      def construct_query(field_pairs, join_with = default_join_with, type = 'field'.freeze)
```
jeremyf: Curious about using freeze on an optional parameter value. Should that be a constant? What is the expected behavior?

jcoyne: it's a compiler optimization. see http://tmm1.net/ruby21-fstrings/

PR #1151, opened 2016-09-16
Cast RDF Literals to strings when indexing.

tpendragon:
Temporary fix until we find a nicer way to index literal information.


tpendragon:
@no-reply Is there a better way to do this casting, or checking if it's a literal?

PR #1152, opened 2016-09-22
Guard for nil logger

jcoyne:


cbeer:
Where does `ActiveFedora::Base.logger` even get defined? Is it possible to just make sure it returns a `NullObject`-logger, if nothing else?

jcoyne:
@cbeer it's set in the railtie here: https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/railtie.rb#L12

It could be a NullObject, but we're guarding for nil everwhere else in the code.

PR #1167, opened 2016-10-28
Add key to changed attributes if type is changed, fixes #1165

### LINKED ISSUE #1165: RDF types are only changeable when the object is first created
Once an object is persisted, it is not possible to change its RDF types.

awead:
Note: I really don't like this solution, but I need to be able to do this.

Ideally, `type` needs to be added to the list of `changed_attributes.keys` but it's buried within ActiveModel and I can't figure out how ActiveFedora is wielding in order to make that happen.

My solution is to "re-query" the resource and compare the existing types with the ones that have been changed in the AF::Base instance.

### DIFF
--- a/lib/active_fedora/persistence.rb
+++ b/lib/active_fedora/persistence.rb
@@ -190,9 +190,17 @@ module ActiveFedora
       end
 
       def execute_sparql_update
-        change_set = ChangeSet.new(self, resource, changed_attributes.keys)
+        changed_attributes_keys = changed_attributes.keys
+        changed_attributes_keys << "type" if type_changed?
+        change_set = ChangeSet.new(self, resource, changed_attributes_keys)
         return true if change_set.empty?
         ActiveFedora.fedora.ldp_resource_service.update(change_set, self.class, id)
+      end
+
+      def type_changed?
+        current = build_ldp_resource(id)
+        current_types = self.class.resource_class.new(current.graph.rdf_subject, data: current.graph.graph.data).type
+        (type.map(&:to_s) - current_types.map(&:to_s)).any?
       end
 
       # Override to tie in an ID minting service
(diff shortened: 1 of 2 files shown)


jcoyne:
This is not a good way. It'll slow down everything

awead:
@jcoyne can you offer another solution? I'm stumped.

jcoyne:
@awead override the type= setter and declare type_will_change!

awead:
@jcoyne yeah, that's what I first tried. The problem is `type_will_change` returns `NoMethodError` because it's not one of the "registered" attributes or properties. Digging into that pulled in the ActiveModel inclusions and I couldn't figure out how that was happening. Any RDF term that's defined using `property` in the model automatically gets those `_will_change` methods created for them, but it's not happening for type.

jcoyne:
Have you considered removing this delegation and changing it to an explicit property? https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/fedora_attributes.rb#L13

jeremyf:
@awead Ping

awead:
@jeremyf I've been trying to rework this, but without success. We can close it, if you want.

jeremyf:
@awead I'm going to spend a bit of time on this today. If you have any additional false steps, pitfalls, or guidance that'd be great.

awead:
@jeremyf the key, I think, is getting `type` to be a registered attribute the way the rest of the property-type attributes are. This involves digging into the way AF is mimicking ActiveRecord, and that's were I started getting lost. Feel free to ping me on Slack if you have specific questions and I'll try to help out as best I can.

jeremyf:
@awead And like that…my time has dissolved. I won't be able to get back to this until January.

no-reply:
I'm closing this as stale. A bugfix for #1165 is still desired.

PR #1181, opened 2016-12-19
Removing invalid namespaces from tests

escowles:
* Removing invalid namespaces from test ids that causes test failures in Fedora 4.7.1-RC-1


jcoyne:
breaking change in Fcrepo on a point release?

escowles:
@jcoyne I've noted the test failures and left a comment in the (release testing page](https://wiki.duraspace.org/display/FF/Release+Testing+-+4.7.1?focusedCommentId=79790774#comment-79790774) — I'm not sure who's around to discuss it before the end of the year, but I suspect we'll try to figure out why that change happened and revert it if possible.

PR #1209, opened 2017-03-05
When updating triples avoid the cache

jcoyne:
Get a fresh copy from Fedora.

Fixes https://github.com/projecthydra/hydra-head/issues/403
Fixes https://github.com/projecthydra/sufia/issues/3065


hortongn:
I pointed Sufia to this branch of AF and tested that this PR does properly fix https://github.com/projecthydra/sufia/issues/3065

merging this

Issue #768, opened 2015-04-14
<< on properties doesn't persist


tpendragon:
We may not support <<, but right now it looks like it works and then doesn't persist.

``` ruby
class Test < ActiveFedora::Base
  property :title, :predicate => RDF::DC.title
end
t = Test.new
t.title << "Test"
t.title # => ["Test"]
t.save
t.title # => []
t.title += ["Test"]
t.title # => ["Test"]
t.save
t.title # => ["Test"]
```

awead:
There's a duplicate of this issue in #540. We also added a test that should still be pending.

atz:
Closing as duplicate.

PR #1212, opened 2017-03-07
Leverage callback in ActiveTriples to mark attributes as dirty

### LINKED ISSUE #540: Change tracking not working when using shift operator on properties?
8) GenericFile delegations that have been saved should be able to be added to w/o unexpected graph behavior
     Failure/Error: expect(f.title).to include("Newer work")
       expected ["New work"] to include "Newer work"
     # ./spec/models/generic_file_spec.rb:304:in `block (4 levels) in <top (required)>'

mbarnett:
Resolves #540,  #768, and possibly other duplicates.

This works hand-in-hand with a separate PR against ActiveTriples (https://github.com/ActiveTriples/ActiveTriples/pull/258). The net result is that ActiveFedora objects have properties properly marked as dirty in circumstances when they previously weren't (most notably, when using `<<` to append to a multi-valued property), so that those changes get persisted instead of silently dropped.

### DIFF
--- a/lib/active_fedora/attribute_methods/dirty.rb
+++ b/lib/active_fedora/attribute_methods/dirty.rb
@@ -3,12 +3,10 @@ module ActiveFedora
     module Dirty
       extend ActiveSupport::Concern
 
-      def set_value(*val)
-        attribute = val.first
+      def dirty_attribute(attribute, values)
         unless [:has_model, :modified_date].include? attribute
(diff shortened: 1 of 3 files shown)


mbarnett:
Naturally, those weren't failing locally...

no-reply:
Closing this, since the corresponding AT changes were replaced by an Observer.

PR #1215, opened 2017-03-20
Empty out nodes directly.

tpendragon:
Without this, ActiveTriples can potentially get stuck in a never-ending
loop.


jcoyne:
Is there any way to create a test for this?

tpendragon:
@jcoyne I wish. I have no idea what even causes this, I just managed to reproduce by uploading enough items in plum and adding `binding.pry` in the appropriate places.

PR #1223, opened 2017-03-29
Merge Solrizer into ActiveFedora

jcoyne:


jeremyf:
I need a bit of background on this. I assumed that merging Solrizer into ActiveFedora would mean that the solrizer gem would no longer be a dependency.

Is that the eventual goal?

jcoyne:
@jeremyf yes, that's the goal. Merge this and solrizer goes to projecthydra-deprecated 👼 . Solrizer does very little (see the diff on this PR) and isn't used by anything that isn't using ActiveFedora.  Of course this is for a major version, so it might be best to wait on  a merge.

awead:
I'm 👍 but I think we should merge the other outstanding PRs, cut a new 11.x release, then merge this and cut AF 12.

darrenleeweber:
This could be un-blocked now that 11.2.0 is cut
- https://github.com/projecthydra/active_fedora/releases/tag/v11.2.0

atz:
Real easy rebase, just a couple lines of conflict.

jeremyf:
@jcoyne I'll go ahead and rebase.

atz:
While I support this merge, I do so mainly towards the end of eradicating runtime use of `solr_name` (i.e. Solrizer) altogether.  `:facetable` is a lie.  `:searchable` is a lie.  `:symbol` is a lie.  

The idea that you could have a repository running and indexed for 20 Million records and some dynamic change in dependency could silently and **usefully** alter what field your application looks for, is bogus.  Once you have indexed data, you should always look to the same place.  Therefore, having the fieldname literal in your application is the only correct path.  IF `solr_name` wants to live on as a developer aid ("what field name should I use for 'title'?"), that's fine by me.  But it needs to be purged from our apps.

PR #1226, opened 2017-04-06
Don't shadow attr_accessor with custom accessor method

jcoyne:


lib/active_fedora/rdf/field_map_entry.rb
```diff
@@ -15,14 +16,15 @@ def initialize
     # Merges any existing values for solr fields with new, incoming values and ensures that resulting values are unique.
     # @param [Symbol] type the data type for the field such as :string, :date, :integer
     # @param [Array] behaviors Solrizer's behaviors for indexing such as :stored_searhable, :symbol
-    # @param [Array] values existing values for the solr field
-    def merge!(type, behaviors, values)
+    # @param [Array] new_values existing values for the solr field
```
cbeer: Does the comment also need to be updated? Is there an even better name for this argument? (`new_values_to_add` or something like that?)

jcoyne: @cbeer can you explain why the comment would need to change?  All I've done here is change the name so that it's easier to read.

cbeer: Did the comment ever make sense?

jcoyne: @awead pinging you as you were the original author ( https://github.com/projecthydra/active_fedora/commit/4d46e30dbdde378e1d480265b0f985a34fd0be30).  I think the word "existing" is confusing and could be eliminated. How about "values to add to the solr field"?

awead: @jcoyne yes, it does seem that `values` are existing values, but I don't think that's what is meant. The existing values are already there, and `values` are what's being added to them.

jcoyne: @awead So, should I change it to this?
```
# @param [Array] new_values values to append into the existing solr field
```

awead: yep, sounds good!

jcoyne:
@cbeer @awead change made.

Issue #1156, opened 2016-10-08
Update external content handling for Fedora 4.7.0-RC


escowles:
Fedora 4.6 returned a 200 OK response when doing a HEAD on an external content resource, but [4.7.0-RC-1]((https://github.com/fcrepo4/fcrepo4/releases/tag/fcrepo-4.7.0-RC-1) returns a 307 Temporary Redirect instead.

This change results in two test failures (error messages below), so we should update to handle the 307 response code:

```
  1) ActiveFedora::File with a sub-resource a binary file streaming the response when there are more than 3 requests because of redirects raises a HTTP redirect too deep Error
     Failure/Error: super

     Ldp::HttpError:
       STATUS: 307 ...
     # ./lib/active_fedora/initializing_connection.rb:22:in `head'
     # ./lib/active_fedora/file.rb:63:in `new_record?'
     # ./lib/active_fedora/with_metadata/metadata_node.rb:24:in `metadata_uri'
     # ./lib/active_fedora/with_metadata/metadata_node.rb:38:in `ldp_source'
     # ./lib/active_fedora/with_metadata/metadata_node.rb:15:in `initialize'
     # ./lib/active_fedora/file.rb:110:in `new'
     # ./lib/active_fedora/file.rb:110:in `metadata'
     # ./lib/active_fedora/file.rb:167:in `block in create_or_update'
     # ./lib/active_fedora/file.rb:166:in `tap'
     # ./lib/active_fedora/file.rb:166:in `create_or_update'
     # ./lib/active_fedora/persistence.rb:29:in `save'
     # ./lib/active_fedora/persistence.rb:237:in `block in save_contained_resources'
     # ./lib/active_fedora/persistence.rb:236:in `each'
     # ./lib/active_fedora/persistence.rb:236:in `save_contained_resources'
     # ./lib/active_fedora/persistence.rb:176:in `_update_record'
     # ./lib/active_fedora/indexing.rb:70:in `_update_record'
     # ./lib/active_fedora/callbacks.rb:244:in `block in _update_record'
     # ./lib/active_fedora/callbacks.rb:244:in `_update_record'
     # ./lib/active_fedora/persistence.rb:159:in `create_or_update'
     # ./lib/active_fedora/callbacks.rb:236:in `block in create_or_update'
     # ./lib/active_fedora/callbacks.rb:236:in `create_or_update'
     # ./lib/active_fedora/persistence.rb:33:in `save!'
[...]

PR #1234, opened 2017-04-20
[WIP] Add support for external files via RFC2017 mime types

cjcolvar:
This PR takes the pattern used by Fedora and some Hydra implementors (see related links below) and enables it at the ActiveFedora::File level.  The use case for Avalon is that we want to store technical metadata and track derivative files that do not live in our repository but somewhere external, possibly a streaming server.  

~~Since the only way to determine if there is external content is to inspect the mime type, a request to fcr:metadata is required to get the file's content.  You can see I had to remove the test in spec/integration/attached_files_spec.rb due to this additional request.~~ Certain calls now require requesting fcr:metadata to get the mime type to check if there is external content. 

I'm guess there are things I've overlooked or edge cases that aren't handled so please give feedback on improvements!

~~The last failing test is related to #1233 and should disappear if I rebase once that is merged.~~

Related tickets and code:
https://github.com/projecthydra/active_fedora/issues/1156
https://github.com/projecthydra/ldp/pull/71
https://github.com/projecthydra/hydra-works/blob/master/lib/hydra/works/services/add_external_file_to_file_set.rb
https://github.com/lbiedinger/ldp/tree/redirects
https://github.com/pulibrary/plum/blob/master/app/jobs/ingest_file_job.rb

### DIFF
--- a/lib/active_fedora/file.rb
+++ b/lib/active_fedora/file.rb
@@ -146,8 +146,7 @@ module ActiveFedora
 
     def external_url
       return nil unless mime_type.start_with? "message/external-body"
-      url = mime_type.split(';').at(2)
-      url.nil? ? nil : url[/\"(.*?)\"/, 1]
+      mime_type[/url=['"](.+?)['"]/i, 1]
     end
 
     def external_url=(external_file_url)
--- a/lib/active_fedora/file.rb
+++ b/lib/active_fedora/file.rb
@@ -107,6 +107,11 @@ module ActiveFedora
       @ds_content ||= retrieve_content
     end
 
+    def external_content
+      return nil unless external_url.start_with?("http")
+      Faraday.get(external_url).body
+    end
+
     def metadata
       @metadata ||= ActiveFedora::WithMetadata::MetadataNode.new(self)
     end
@@ -116,6 +121,7 @@ module ActiveFedora
     end
 
     def content_changed?
+      return false if external_content?
       return true if new_record? && !local_or_remote_content(false).blank?
       local_or_remote_content(false) != @ds_content
     end
@@ -139,6 +145,20 @@ module ActiveFedora
       false
     end
 
+    def external_content?
+      external_url.present?
+    end
+
+    def external_url
+      return nil unless mime_type.start_with? "message/external-body"
+      url = mime_type.split(';').at(2)
+      url.nil? ? nil : url[/\"(.*?)\"/, 1]
+    end
+
+    def external_url=(external_file_url)
+      self.mime_type = "message/external-body; access-type=URL; URL=\"#{external_file_url}\""
+    end
+
     # serializes any changed data into the content field
     def serialize!; end
 
@@ -147,12 +167,13 @@ module ActiveFedora
     end
 
     def content=(string_or_io)
+      raise "This file has external content.  First call external_url=nil if you want to use content=." if external_content?
       content_will_change! unless @content == string_or_io
       @content = string_or_io
     end
 
     def content
-      local_or_remote_content(true)
+      external_content? ? external_content : local_or_remote_content(true)
     end
 
     def self.relation
--- a/lib/active_fedora/file.rb
+++ b/lib/active_fedora/file.rb
@@ -103,13 +103,8 @@ module ActiveFedora
     end
 
     def remote_content
-      return if new_record?
+      return if new_record? && !external_content?
       @ds_content ||= retrieve_content
-    end
-
-    def external_content
-      return nil unless external_url.start_with?("http")
-      Faraday.get(external_url).body
     end
 
     def metadata
@@ -173,7 +168,7 @@ module ActiveFedora
     end
 
     def content
-      external_content? ? external_content : local_or_remote_content(true)
+      local_or_remote_content(true)
     end
 
     def self.relation
@@ -195,7 +190,12 @@ module ActiveFedora
       end
 
       def retrieve_content
-        ldp_source.get.body
+        ldp_response = ldp_source.get
+        if ldp_response.response.status == 307
+          Faraday.get(ldp_response.response["location"])
+        else
+          ldp_response.body
+        end
       end
 
       def ldp_headers
@@ -217,7 +217,7 @@ module ActiveFedora
       end
 
       def local_or_remote_content(ensure_fetch = true)
-        return @content if new_record?
+        return @content if new_record? && !external_content?
 
         @content ||= ensure_fetch ? remote_content : @ds_content
         @content.rewind if behaves_like_io?(@content)
--- a/spec/integration/attached_files_spec.rb
+++ b/spec/integration/attached_files_spec.rb
@@ -1,31 +1,6 @@
 require 'spec_helper'
 
 describe ActiveFedora::AttachedFiles do
-  describe "#has_subresource" do
-    before do
-      class FooHistory < ActiveFedora::Base
-        has_subresource 'child'
-      end
-    end
-    after do
-      Object.send(:remove_const, :FooHistory)
-    end
-
-    context "when the object exists" do
-      let!(:o) { FooHistory.create }
-      before do
-        o.child.content = "HMMM"
-        o.save
-      end
-
-      it "does not need to do a head on the children" do
-        f = FooHistory.find(o.id)
-        expect(f.ldp_source.client).not_to receive(:head)
-        f.child.content
-      end
-    end
-  end
-
   describe "Datastreams synched together" do
     before do
       class DSTest < ActiveFedora::Base
--- a/spec/integration/attached_files_spec.rb
+++ b/spec/integration/attached_files_spec.rb
@@ -1,6 +1,31 @@
 require 'spec_helper'
 
 describe ActiveFedora::AttachedFiles do
+  describe "#has_subresource" do
+    before do
+      class FooHistory < ActiveFedora::Base
+        has_subresource 'child'
+      end
+    end
+    after do
+      Object.send(:remove_const, :FooHistory)
+    end
+
+    context "when the object exists" do
+      let!(:o) { FooHistory.create }
+      before do
+        o.child.content = "HMMM"
+        o.save
+      end
+
+      it "does not need to do a head on the children" do
+        f = FooHistory.find(o.id)
+        expect(f.ldp_source.client).not_to receive(:head)
+        f.child.content
+      end
+    end
+  end
+
   describe "Datastreams synched together" do
     before do
       class DSTest < ActiveFedora::Base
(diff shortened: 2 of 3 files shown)


lib/active_fedora/file.rb
```diff
@@ -147,12 +167,13 @@ def to_solr(solr_doc = {}, _opts = {})
...
 
     def content=(string_or_io)
+      raise "This file has external content.  First call external_url=nil if you want to use content=." if external_content?
       content_will_change! unless @content == string_or_io
       @content = string_or_io
     end
 
     def content
-      local_or_remote_content(true)
+      external_content? ? external_content : local_or_remote_content(true)
```
escowles: Do we need to explicitly retrieve the external content URL from the description, and then retrieve it?  Can't we just follow Fedora's redirect?

lib/active_fedora/file.rb
```diff
@@ -139,6 +145,20 @@ def metadata?
...
     end
 
+    def external_content?
+      external_url.present?
+    end
+
+    def external_url
+      return nil unless mime_type.start_with? "message/external-body"
+      url = mime_type.split(';').at(2)
+      url.nil? ? nil : url[/\"(.*?)\"/, 1]
```
escowles: Is there a HTTP header parser we can use that would be more robust against slight changes to the external-body format?  E.g., I know the spaces are optional, and I think you might be able to use single quotes or change the order of the tokens so the URL was in the second position, etc.

cjcolvar: I couldn't really find anything to do this parsing for us.  Any suggestions?

escowles: I thought one of the HTTP clients would have this (and they must internally), but I couldn't find anything.  I did find a gem called content-type that will do this: https://rubygems.org/gems/content-type

```
ContentType.parse("message/external-body; access-type=\"URL\"; URL=\"http://example.org/1\"").parameters['url']
```

cjcolvar: Is it worth pulling in this gem (of 2 commits from 3 years ago) along with its dependency parslet for this?  Maybe it would be better to come up with something good enough that we own instead.

escowles: Yes, I'm overthinking this — we could probably even just use a regex like:

```
url.match(/url=['"](.+?)['"]/i).captures.first
```

lib/active_fedora/file.rb
```diff
@@ -139,6 +140,19 @@ def metadata?
...
+      external_url.present?
+    end
+
+    def external_url
+      return nil unless mime_type.start_with? "message/external-body"
+      mime_type[/url=['"](.+?)['"]/i, 1]
+    end
+
+    def external_url=(external_file_url)
+      self.mime_type = "message/external-body; access-type=URL; URL=\"#{external_file_url}\""
```
awead: @cjcolvar  This definitely makes it easy. I'm wondering if we could use another property. Could we make content itself an HTML redirect so that when someone hits the resource it actually redirects to the location? Or does that work with using mime_type?

lib/active_fedora/file.rb
```diff
@@ -147,6 +161,7 @@ def to_solr(solr_doc = {}, _opts = {})
     end
 
     def content=(string_or_io)
+      raise "This file has external content.  First call external_url=nil if you want to use content=." if external_content?
```
awead: @cjcolvar might be nice to have a `external_content_reset` method that does the setting to nil for you.

lib/active_fedora/file.rb
cjcolvar: Fedora currently redirects if the file has a mime_type using message/external-body (RFC2017).  This means the content is still stored but is never exposed.

awead: @cjcolvar interesting. so you could still have binary content on `content` but the override to `mime_type` would force a redirect?

escowles: I'd have to check to make sure, but I think you could store content but you wouldn't be able to retrieve it.  But once you changed the mime type to something else, then you could probably retrieve the content again.

cjcolvar: That is currently how `File` and Fedora work.  I'm attempting to have `.content` return the bits for either external or internal.  Thus making the `File` act like it either has external or internal content but never both at the same time.  I had started with a separate `.external_content` method but thought this approach clearer.

afred: So should we consider what makes a resource with external content valid, and enforce it?
 I.e...

A resource with external content is valid iff _all_ of the following are true:
  * mime type begins with `message/exenal-body`
  * mime type contains `access-type=URL;`
  * mime type contains `URL=x`, where `x` is a valid URI.
  * the resource contains no binary content

Conversely, a resource containing internal content is invalid if _any_ of the following are true:
  * the mime type contains `message/external-body`
  * the mime type contains `access-type=URL`
  * the mime type contains `URL=x` where `x` is anything at all.

escowles: If the mime type is `message/external-body` with `access-type=URL` and `URL=anything`, then you can't tell if there is binary content or not (because Fedora will return a redirect instead of the binary).  I guess you could check the SHA1 checksum against the known checksum for an empty binary, but that doesn't seem like a good idea to me.

So I think we should leave the content out and say a resource with external content is valid if all of these are true:
* mime type is `message/external-body`
* the `access-type` parameter is present and the value is `URL`
* the `URL` parameter is present and the value is a valid URL

I'm less sure about the other stuff: `message/external-body` can handle a couple of other URL schemes (ftp, sftp, local-file), and in theory I could imagine someone wanting to use that to point to those.  So I think it would be better to be flexible on that.

afred: When setting external content via `#external_content=`, would it make sense to first set the content to a 0 byte string? The purpose would be to avoid dangling references to binary data.

lib/active_fedora/file.rb
```diff
@@ -196,7 +216,7 @@ def uploaded_file?(payload)
       end
 
       def local_or_remote_content(ensure_fetch = true)
```
afred: Would it make sense to settle on the word "external" instead of "remote" for this method (and any others that may be applicable)? We could `alias :local_or_remote_content, :local_or_external_content`, and add a deprecation warning.

cjcolvar: My understanding is that remote_content here means the version of the content that lives in fedora as opposed to the local yet to be persisted version of the content.  External content is different from those two and thus a new name.  It might be appropriate to call this method `:local_or_remote_or_external_content` now but that seems wordy and confusing.

lib/active_fedora/file.rb
cjcolvar: That sounds like a good idea to me.

awead: @escowles why both `access-type` and `URL`? Not disagreeing, just wondering about the need for redundancy. Fedora 3 also supported things like nfs-mounted files for its external datastreams. It would be nice if can offer the same flexibility here too.

escowles: @awead I'm mostly thinking about someone having `message/external-body; access-type="local-file"; local-file="/path/to/binary"` — I'm not sure what we should do with that, and I think it would be best to be tolerant in what we allow people to do until we have a better handle on that.

awead: @escowles I would think a local file could use the file uri such as `message/external-body; access-type=URL; URL="file:///path/to/binary.bin"` but I tried that in Fedora and it doesn't seem to work. Could we limit it urls for the time being?

awead: FYI, tried local-file in Fedora too and no dice.

cjcolvar: @awead Can you explain how it didn't work?  Do you mean wouldn't persist, downloading content doesn't give you a redirect, or something else?

awead: I updated the mime_type in Fedora directly via SPARQL. Using `access_type=URL` with `URL=http://[etc]` worked fine. When I clicked "Download content" in the Fedora UI, it redirected to the content. Note, it actually didn't "download" it, it presented the image in the window, which is a true redirect.

When I used either `URL=file:///` to a local path, or @escowles `local-file` option, neither worked. I clicked on "Download content" and nothing happened. I didn't look at logs, though. But my point was that maybe we should just support a basic HTTP protocol first, and work on others later.

escowles: @awead Right — the current Fedora impl will only send a redirect for `access-type="URL"` and a HTTP URL.  With any other access-type, it'll just serve the internal binary (0-byte or otherwise).

But this behavior is under active discussion in the Fedora API spec group, so I wouldn't want to bake anything else in right now.  So my proposal is:

* If `content-type` starts with `message/external-body`, the `access-type` parameter is `URL` and the `URL` parameter contains something that looks like a URL, then treat it as external content.
* Everything else is treated as a normal binary.

When using `file:///` URLs and/or `access-type="local-file"` get specified, then we can add support for those as needed.

awead: @escowles oh, sorry, I misunderstood. I thought you wanted to support those options now. 👍 then to above.

cjcolvar: @escowles Are you suggesting if I want to create an AF::File that points to a non-resolvable or non-HTTP URL then I might be better off storing that in another field (like `ebucore:locator`) and setting empty content instead of using `message/external-body` in the `mime_type`?

I figured it would be better to use the same approach to store the external reference no matter the protocol or resolvability and mark the `AF::File` as external content in all of those cases because that is what it is representing.

escowles: @cjcolvar I don't know about this.  Right now, Fedora sends redirects, but there is a lot of interest in having Fedora proxy content instead (both HTTP and potentially local files too).  My main concern is that if we support anything beyond the currently-implemented behavior (which is already in use by several apps), it will complicate implementing whatever gets decided about proxying or supporting files instead of just HTTP.  I'm hoping to discuss this with the new Fedora API spec group that's meeting tomorrow.

cjcolvar: @escowles Great!  Looking forward to the outcome of that then!

cjcolvar: @escowles Any updates from the Fedora API spec group meeting?

escowles: The current proposal for the API spec is here: https://github.com/fcrepo/fcrepo-specification/wiki/External-Content (this wiki page has been generally agreed upon and will be incorporated into the spec shortly).

The basic pattern is different from the current implementation:
* Either `url` or `local-file` are supported, and they are to be interpreted by the server.
* The server will then proxy the content and the client doesn't need to know anything about the content being external.
* The location of the external content will be exposed to the client in a `Content-Location` header.
* Supporting redirects like the current implementation is still undecided — if anyone has a use case for redirects that proxying won't cover, I'd recommend opening an issue in https://github.com/fcrepo/fcrepo-specification/issues

So, I think we should support the current behavior now, but I don't think we need to do anything about `local-file` or other types right now.

cjcolvar:
Closing because I'm no longer working on this and might be irrelevant now.

PR #1235, opened 2017-04-20
Metadata node save

cjcolvar:
This PR builds on #1233 and fixes the resetting of metadata_uri so if you call save on the metadata node it will reinitialize with the described_by uri from it's parent file.


awead:
@cjcolvar I think this needs a rebase. The first commit should already be in master.

cjcolvar:
@awead I rebased this so it should be ready to go now.

PR #1249, opened 2017-05-22
Make solr.yml consistent w/ blacklight.yml (and itself)

atz:
- honor SOLR_URL in all 3 environments
- Use 127.0.0.1 for both dev and test
- fix copy/paste error repetition of SOLR_TEST_PORT in dev environment

See also: 
- https://github.com/projecthydra-labs/hyrax/issues/1016
- https://github.com/projecthydra/hydra-head/pull/411


mjgiarlo:
I wonder if we should use `localhost` instead of `127.0.0.1`, though I'm not going to hold this PR up over that since it's unrelated to the suggested change.

atz:
Open question.  Don't really care, as long as we are consistent.

PR #1252, opened 2017-05-30
File metadata should have a modified_date

cjcolvar:
ActiveFedora::Base objects have `.modified_date` alongside a `.create_date`, but ActiveFedora::File objects only have a `.create_date`.  This PR adds a `.modified_date` to match and make it easier to get the last modified date for a NonRdfSource.

### DIFF
--- a/lib/active_fedora/file/attributes.rb
+++ b/lib/active_fedora/file/attributes.rb
@@ -49,6 +49,11 @@ module ActiveFedora::File::Attributes
     created && created.first
   end
 
+  def modified_date
+    modified = metadata.attributes["http://fedora.info/definitions/v4/repository#lastModified"]
+    modified && modified.first
+  end
+
   private
 
     # Fcrepo4.digest was used by Fedora < 4.3, but it was removed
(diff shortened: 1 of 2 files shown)


lib/active_fedora/file/attributes.rb
```diff
@@ -49,6 +49,11 @@ def create_date
     created && created.first
   end
 
+  def modified_date
+    modified = metadata.attributes["http://fedora.info/definitions/v4/repository#lastModified"]
+    modified && modified.first
```
atz: If there are multiple values here, mightn't we expect the most recent rather than just the first?

jcoyne: @atz would fedora ever return more than one value for lastModified? I've never seen this before.

cjcolvar: I wouldn't think it would be possible except potentially in "relaxed mode" coming in 4.7.4 as mentioned here: https://wiki.duraspace.org/display/FEDORA4x/How+to+allow+user-updates+to+certain+server+managed+triples .
@mikedurbin is it ever possible for fedora:lastModified to be multivalued?

mikedurbin: No, it should never be multi-valued, even in relaxed mode.

PR #1257, opened 2017-06-12
No need to set RSolr 2.0 explicitly

jcoyne:
2.0.2 is the latest and will be installed by default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
   include:
     - rvm: 2.3.4
       env: "RAILS_VERSION=4.2.8"
-    - env: "RSOLR_VERSION=2.0.1"
+    - env: "RAILS_VERSION=5.1.1"
 global_env:
   - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
 before_install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -6,7 +6,7 @@ matrix:
   include:
     - rvm: 2.3.4
       env: "RAILS_VERSION=4.2.8"
-    - env: "RSOLR_VERSION=2.0.1"
+    - env: "RAILS_VERSION=5.1.1"
```
dunn: Should adding Rails 5 to the matrix be a separate commit?

jcoyne: Removing the RSOLR_VERSION makes this build only have a single job.  Prior to removing the RSOLR_VERSION it was building on rails 5.1.1 (latest) and 4.2.8.

jcoyne:
@cjcolvar YAGNI? Do you have a use case that needs rsolr 1?

.travis.yml
dunn: Gotcha.

cjcolvar:
No, my point was more that we might want to test targeting a specific rsolr version in the future which seems like was the use case for adding this in the first place.  But I don't have a specific need so I'm fine with removing it and we can always add it back in later if we do need it.

PR #1271, opened 2017-07-25
Request options allow for setting timeout of Fedora client

cjcolvar:
When attempting to run reindex_everything, my Fedora instance sits in silence for 10+ minutes before returning the first bytes of a response to a get request of the base path.  Faraday throws a timeout after a few seconds so I need to be able to configure it to allow for these abnormally long periods of waiting.


cjcolvar:
Fixed #1105.

PR #1274, opened 2017-07-27
Raise a helpful error when the document has no model

jcoyne:


cbeer:
Do you plan to forward-port this as needed to 7.x and 8.x?

jcoyne:
Yes, "as needed"

PR #1277, opened 2017-08-19
Prevent `exists?` from throwing exceptions

### LINKED ISSUE #1276: exists? throws exception on class mismatch instead of returning false
Example problem:
```ruby
> ActiveFedora::Base.exists? "wg827ks1643"
true
> Collection.exists? "wg827ks1643"
true
> GenericWork.exists? "wg827ks1643"
*** ActiveFedora::ActiveFedoraError Exception: Model mismatch. Expected GenericWork. Got: Collection
```
This doesn't fit the intended method sig/behavior of `exists?` at all. I'm not asking if the ID is taken *anywhere* (in which case I would ask `ActiveFedora::Base` or my minter). I'm asking: "Does an object _of this class_ exist with the given ID?" This should be obvious by the fact I am passing the message to the class itself.  If the object does not match the class, then the response should be `false`. 

The answer should **never** be an Exception, because if you can't safely ask whether a thing exists, the rest of your program logic is going to get ornate or fall apart.

Implementation is here:
https://github.com/samvera/active_fedora/blob/master/lib/active_fedora/relation/finder_methods.rb#L78-L91

Fuller trace:
```
ActiveFedora::ActiveFedoraError: Model mismatch. Expected GenericWork. Got: Collection
  /Users/atz/.rvm/gems/ruby-2.4.1/gems/active-fedora-11.4.0/lib/active_fedora/relation/finder_methods.rb:210:in `class_to_load'
  /Users/atz/.rvm/gems/ruby-2.4.1/gems/active-fedora-11.4.0/lib/active_fedora/relation/finder_methods.rb:196:in `load_from_fedora'
  /Users/atz/.rvm/gems/ruby-2.4.1/gems/active-fedora-11.4.0/lib/active_fedora/relation/finder_methods.rb:181:in `find_one'
  /Users/atz/.rvm/gems/ruby-2.4.1/gems/active-fedora-11.4.0/lib/active_fedora/relation/finder_methods.rb:234:in `find_with_ids'
  /Users/atz/.rvm/gems/ruby-2.4.1/gems/active-fedora-11.4.0/lib/active_fedora/relation/finder_methods.rb:52:in `find'
  /Users/atz/.rvm/gems/ruby-2.4.1/gems/active-fedora-11.4.0/lib/active_fedora/relation/finder_methods.rb:85:in `exists?'
```
As a related problem, `exists?` should not rely on fully loading the object from fedora (`load_from_fedora`) either.  That result is going to get thrown out.  If I wanted the cost of fully loading object and the risk of an exception, I would just call `find` myself!

atz:
Fixes #1276 

Resolve failing test with a distinct error class, allowing `exists?` to catch the new error as well as
`ObjectNotFoundError`, returning `false`.

- Refactored tests to reuse ad hoc classes
- added negative tests for `#find`
- Consolidated single-line single-use private methods

### DIFF
--- a/lib/active_fedora/errors.rb
+++ b/lib/active_fedora/errors.rb
@@ -40,6 +40,10 @@ module ActiveFedora #:nodoc:
   class AssociationNotFoundError < ConfigurationError #:nodoc:
   end
 
+  # Raised when an object is loaded from Fedora by an incompatible class
+  class ModelMismatch < ActiveFedoraError
+  end
+
   # This error is raised when trying to destroy a parent instance in N:1 or 1:1 associations
   # (has_many, has_one) when there is at least 1 child associated instance.
   # ex: if @project.tasks.size > 0, DeleteRestrictionError will be raised when trying to destroy @project
--- a/lib/active_fedora/ldp_resource.rb
+++ b/lib/active_fedora/ldp_resource.rb
@@ -18,8 +18,7 @@ module ActiveFedora
     end
 
     def marshal_load(data)
-      ivars = data
-      ivars.each { |name, val| instance_variable_set(name, val) }
+      data.each { |name, val| instance_variable_set(name, val) }
     end
   end
 end
--- a/lib/active_fedora/relation/finder_methods.rb
+++ b/lib/active_fedora/relation/finder_methods.rb
@@ -1,7 +1,6 @@
 module ActiveFedora
   module FinderMethods
-    # Returns the first records that was found.
-    #
+    # Returns the first record that was found.
     # @example
     #  Person.where(name_t: 'Jones').first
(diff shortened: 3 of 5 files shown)


atz:
@no-reply correct, I did not drill down or alter that behavior in the scope of this ticket/PR.  @cbeer pointed out in #1278 that most of that module was copied directly from rails, but I note we have a couple important differences from activerecord.  

AR has the advantage of knowing the "type" info ahead of time and being able to ask the primary key on a specific table (index) is sufficient to answer `exists?(id)`.  We don't have that.  We need to retrieve ALL the model information to determine whether a record matches the object class.  (Is that info in HEAD?) `ActiveFedora::Base` is the exception, where it *doesn't* need model info and certainly could use HEAD.

PR #1282, opened 2017-10-09
Adding update_index callback hooks

jeremyf:
To implement nested indexing of objects, we need access to a callback
hook (and one that we can then trigger an async job for). This makes
the callbacks created in Hyrax a first class citizen.

Extracting from samvera/hyrax#1736


mjgiarlo:
@jcoyne @jeremyf Build running again after rebase. Feel free to merge when this goes :green_heart:

PR #1286, opened 2017-11-06
Treat `RDF::Term` implementers as singular values

no-reply:
Some `RDF::Term` objects implement `#each`, including `ActiveTriples::RDFSources`. The old checks for singular fail on simple cases like `work.single = RDF::Node.new; work.single = work.single`. The new check has a special case for terms.

Ideally, we could check something like `#to_ary`, but behavior is not consistent across the relevant dependency libraries. E.g. `ActiveTriples::Relation#to_ary` is unimplemented, while `RDF::Queryable#to_ary` is (the opposite would be better for our use).

### DIFF
--- a/lib/active_fedora/attributes/property_builder.rb
+++ b/lib/active_fedora/attributes/property_builder.rb
@@ -13,9 +13,10 @@ module ActiveFedora::Attributes
     end
 
     def self.define_writers(mixin, name)
+      # RDF Terms are singular, even if they respond to `#each`
       mixin.class_eval <<-CODE, __FILE__, __LINE__ + 1
         def #{name}=(value)
-          if value.present? && !value.respond_to?(:each)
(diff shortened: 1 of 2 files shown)


lib/active_fedora/attributes/property_builder.rb
```diff
@@ -43,9 +44,10 @@ def #{name}_id(*args)
     end
 
     def self.define_singular_writers(mixin, name)
+      # RDF Terms are singular, even if the respond to `#each`
```
jcoyne: should `the` be `they`?

no-reply: I suppose so.

no-reply: I force pushed the typo fix

jcoyne: Not seeing it.

no-reply: I missed that I had this in two places.

PR #1291, opened 2017-11-10
Update build matrix

no-reply:
The proposed updates are:

  - Build latest of Ruby 2.3, 2.4, and ruby-head.
  - Keep Ruby 2.3.4.
  - Allow failure on ruby-head (this is a canary).
  - Use `trusty` explicitly (this provides stability for the default).

In addition to these changes, I would suggest adding a weekly cron job to the travis configuration to build `master` if a build has not occurred in the last 24 hours.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,19 @@
 language: ruby
 cache: bundler
+dist: trusty
 sudo: false
-rvm: 2.4.1
+
+rvm:
+  - 2.3
+  - 2.3.4
+  - 2.4
+  - ruby-head
+
 matrix:
+  allow_failures: ruby-head
   include:
-    - rvm: 2.3.4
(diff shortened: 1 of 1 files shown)


jcoyne:
Looks like apache is cutting us off on hitting their svn server. Maybe we need to put the solr version to use in solr_wrapper.  Cron build seems like a good idea 👍

.travis.yml
```diff
@@ -1,12 +1,19 @@
 language: ruby
 cache: bundler
+dist: trusty
 sudo: false
-rvm: 2.4.1
+
+rvm:
+  - 2.3
```
jcoyne: Why are we testing 2.3 and 2.3.4. It looks like when you ask for 2.3, that Travis creates an alias to 2.3.4:
```
rvm alias create 2.3 ruby-2.3.4
```

PR #1294, opened 2017-11-15
Rename solr/config directory to solr/conf

jcoyne:
This matches how the config is distributed with Solr and
how Blacklight creates its configs.  Having only one way to do this
seems like it would reduce confusion.


jrgriffiniii:
> This looks ok, but hydra-head also has a config directory. Does this conflict?

I'm not certain if this is a blocking obstacle, but this does indeed seem to conflict in at least one location within `hydra-head`.  Please see https://github.com/samvera/hydra-head/blob/master/hydra-access-controls/tasks/hydra-access-controls.rake#L9

cjcolvar:
I wonder if we could get around this by using `with_test_server` in hydra-head and avoid specifying the directory.  This could clear the way for this PR then.

tpendragon:
CCMWG thinks we can fix this here, and then go and fix https://github.com/samvera/hydra-head/issues/485

jrgriffiniii:
I've rebased this and added a commit which updates the CircleCI configuration to use the new Solr configuration directory paths.

PR #1295, opened 2017-11-22
Allow for a pluggable minter service

jcoyne:


lib/active_fedora/persistence.rb
```diff
@@ -2,6 +2,13 @@ module ActiveFedora
   # = Active Fedora Persistence
   module Persistence
     extend ActiveSupport::Concern
+    extend ActiveSupport::Autoload
+    autoload :NullIdentifierService
+
+    included do
+      class_attribute :identifier_service_class
+      self.identifier_service_class = NullIdentifierService
```
cam156: @jcoyne could this class be read from the configuration?

jenlindner: Is that a blocker for you, @cam156?

awead: using class attributes is consistent with what we're doing elsewhere in AF, so I'm 👍  for this.

PR #1299, opened 2018-01-12
Bump version to 12.0.1

jcoyne:


no-reply:
This version was released January 12, 2018.

PR #1302, opened 2018-03-12
Prefer om_term_values to send when solrizing within SimpleDatastream

mjgiarlo:
This allows downstream users to use fields that conflict with public AF methods. Case in point: I need to support a field named "prefix" which entered the public API around AF 7.0.0.

### DIFF
--- a/lib/active_fedora/simple_datastream.rb
+++ b/lib/active_fedora/simple_datastream.rb
@@ -1,12 +1,12 @@
 module ActiveFedora
-  #This class represents a simple xml datastream. 
+  #This class represents a simple xml datastream.
   class SimpleDatastream < OmDatastream
 
     class_attribute :class_fields
     attr_accessor :fields
     self.class_fields = []
-    
-    
+
+
      set_terminology do |t|
        t.root(:path=>"fields", :xmlns=>nil)
      end
@@ -16,7 +16,7 @@ module ActiveFedora
         xml.text(name)
       end
     end
-    
+
 
     #Constructor. this class will call self.field for each DCTERM. In short, all DCTERMS fields will already exist
     #when this method returns. Each term is marked as a multivalue string.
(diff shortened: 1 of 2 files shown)


lib/active_fedora/simple_datastream.rb
```diff
@@ -75,25 +75,20 @@ def update_indexed_attributes(params={}, opts={})
...
     def to_solr(solr_doc = Hash.new) # :nodoc:
       @fields.each do |field_key, field_info|
         next if field_key == :location ## FIXME HYDRA-825
-        things = send(field_key)
-        if things 
-          field_symbol = ActiveFedora::SolrService.solr_name(field_key, type: field_info[:type])
-          things.val.each do |val|    
-            ::Solrizer::Extractor.insert_solr_field_value(solr_doc, field_symbol, val.to_s )         
-          end
+        field_symbol = ActiveFedora::SolrService.solr_name(field_key, type: field_info[:type])
```
mjgiarlo: Moved this up. No need to invoke it once per iteration within `#each`, I don't think.

lib/active_fedora/simple_datastream.rb
```diff
@@ -75,25 +75,20 @@ def update_indexed_attributes(params={}, opts={})
...
+        om_term_values(field_key).each do |val|
```
mjgiarlo: `#om_term_values` always returns an Array, so don't bother with the conditional or the variable assignment.

lib/active_fedora/simple_datastream.rb
```diff
@@ -75,25 +75,20 @@ def update_indexed_attributes(params={}, opts={})
...
+          ::Solrizer::Extractor.insert_solr_field_value(solr_doc, field_symbol, val.to_s)
         end
       end
-      return solr_doc
+      solr_doc
```
mjgiarlo: Unnecessary `return` usage.

lib/active_fedora/simple_datastream.rb
```diff
@@ -75,25 +75,20 @@ def update_indexed_attributes(params={}, opts={})
...
+      Nokogiri::XML::Document.parse("<fields/>")
     end
 
     def to_solr(solr_doc = Hash.new) # :nodoc:
       @fields.each do |field_key, field_info|
         next if field_key == :location ## FIXME HYDRA-825
-        things = send(field_key)
-        if things 
-          field_symbol = ActiveFedora::SolrService.solr_name(field_key, type: field_info[:type])
-          things.val.each do |val|
```
mjgiarlo: I couldn't find any evidence that `things` would return a thing that responds to `:val`. Help?

mjgiarlo:
Would y'all mind having a look at this? @cbeer @jcoyne @dchandekstark

mjgiarlo:
:point_up: also @no-reply @cjcolvar @mbklein

lib/active_fedora/simple_datastream.rb
jcoyne: ¯\_(ツ)_/¯

mjgiarlo:
@cjcolvar Done! Mind taking another look?

PR #1315, opened 2018-05-30
Allow property redefinition on `MetadataNode`

no-reply:
In order to change `MetadataNode`'s `:mime_type` property, we need to be able to
call `MetadataNode.property` without a `.parent_class` set. This isn't possible
normally because `.property` is overridden after the initial definition of
`:mime_type` to patch in delegation to the parent class when a new property is
defined. That delegation fails when `.parent_class` is `nil`, as is the case on
the base `MetadataNode` class.

We now only do that delegation if `.parent_class` is defined.


no-reply:
> Can you add a test when the parent class does exist and one where it doesn't?

That seems like a nice thing to add, but I'm not inclined to put it in this PR. This method override was previously untested, the use of the module is more than a little byzantine to me, and the current code seems to improve the state of things.

I could open a tech debt ticket, instead?

jcoyne:
I think that nobody is ever going to understand this code more than you do right now. Since this code is byzantine, we really need it to be tested. So I'd really appreciate a second test. 🙏

no-reply:
just closing this as unlikely to get merged.

PR #1322, opened 2018-07-10
Run and test validation callbacks

cjcolvar:
As reported various times (https://github.com/samvera/active_fedora/issues/914, https://github.com/samvera/active_fedora/pull/797, https://github.com/avalonmediasystem/avalon/pull/2984), it looks like validation callbacks aren't running.  

This PR adds a test case for before and after validation callbacks and then copies the code used in ActiveModel: https://github.com/rails/rails/blob/master/activemodel/lib/active_model/validations/callbacks.rb#L114-L119

### DIFF
--- a/lib/active_fedora/validations.rb
+++ b/lib/active_fedora/validations.rb
@@ -92,5 +92,12 @@ module ActiveFedora
       def perform_validations(options = {}) # :nodoc:
         options[:validate] == false || valid?(options[:context])
       end
+
+    private
+
+      # Overwrite run validations to include callbacks.
+      def run_validations!
+        _run_validation_callbacks { super }
+      end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/active_fedora/validations.rb
```diff
@@ -92,5 +92,12 @@ def raise_validation_error
       def perform_validations(options = {}) # :nodoc:
         options[:validate] == false || valid?(options[:context])
       end
+
+    private
+
+      # Overwrite run validations to include callbacks.
```
jcoyne: I'm not sure I understand this comment. Can you be more verbose?

jcoyne: Never mind. I see that is a copy from AM.

cjcolvar:
@jcoyne I bumped the `solr_wrapper` major version to fix the travis build.  Does this change look ok or do you think it should be `>= 1.0, < 3.0` instead of `~> 2.0`?

PR #1365, opened 2018-09-26
Prepare release 12.1.0

no-reply:
Release notes: https://github.com/samvera/active_fedora/releases/tag/untagged-fe547feba4f09791e070


no-reply:
I'm very skeptical that coveraged decreased on this change. Thoughts?

PR #1369, opened 2018-11-01
Remove deprecated Solr StandardFilter

jcoyne:
This filter is deprecated and according to the solr docs it is
inoperative if "luceneMatchVersion (in solrconfig.xml) is higher than
3.1."  The luceneMatchVersion in active_fedora is currently 5.0.0

See https://lucene.apache.org/solr/guide/7_0/filter-descriptions.html#FilterDescriptions-StandardFilter


cjcolvar:
Closed by https://github.com/samvera/active_fedora/pull/1380

PR #1375, opened 2019-01-23
Add and test Ruby 2.6.0

no-reply:
Adding 2.6 to the build matrix for `master`.


jcoyne:
The Ruby 2.3 build has installed Rails 0.9.5. 😖

PR #1376, opened 2019-01-23
Remove references to `LatLonType` from Solr schema

### LINKED ISSUE #1370: Solr config refers to deprecated LatLonType
```
    <fieldType name="location" class="solr.LatLonType" subFieldSuffix="_coordinate"/>
```

I don't think we're using this anywhere. Can we drop this field type or do we want to migrate to one of: `LatLonPointSpatialField`, `SpatialRecursivePrefixTreeField`, or `RptWithGeometrySpatialField`?

no-reply:
This type is deprecated in Solr. Its removal from the template here shouldn't
affect existing adopters who may be using it. It will be removed only for folks
running the generator after this commit.

Fixes #1370


cjcolvar:
@no-reply What do you think about incorporating @jcoyne 's suggestion?

bess:
@jcoyne After conversations with @cjcolvar about this, we put your suggestion into a separate issue, here: https://github.com/samvera/active_fedora/issues/1381

PR #1379, opened 2019-03-21
Use samvera orb

cjcolvar:


spec/spec_helper.rb
```diff
@@ -20,7 +20,6 @@
 require 'rspec/its'
 require 'equivalent-xml/rspec_matchers'
 require 'logger'
-require 'pry' unless ENV['TRAVIS']
```
no-reply: we probably still want this, right? just `require 'pry'`, now?

tpendragon: Or even just `unless ENV['CI']`

PR #1389, opened 2019-06-26
core reload rescue edge case

### LINKED ISSUE #1388: reload on object copy results in uncaught error
When two copies of an object are live and one is destroyed and reload is called with the other, an uncaught Ldp::Gone error is raised as the destroyed flag has not been set on the other copy.  The expected behavior when reload is called on a destroyed object is to raise an ActiveFedora::ObjectNotFoundError.
This scenario can occur in Hyrax when the wings module is used; in the persister delete method, a copy of the object is destroyed, and the destroyed flag is not set on the active copy.

lsat12357:
Fixes #1388 
This adds a rescue block to the reload method in core; it handles the case where two copies of an object are live, one is destroyed, but reload is called the other which will not have the destroyed flag set, resulting in uncaught Ldp::Gone, rather than the expected ActiveFedora::ObjectNotFoundError.
This scenario can occur in Hyrax when the wings module is in use; a monkey patch by @no-reply to temporarily handle this in Hyrax is in progress.

### DIFF
--- a/lib/active_fedora/core.rb
+++ b/lib/active_fedora/core.rb
@@ -62,6 +62,9 @@ module ActiveFedora
       refresh
       load_attached_files
       self
+    rescue Ldp::Gone => err
+      logger.error("Tried to reload an object that has been destroyed.\n\t#{err.message}")
+      raise ActiveFedora::ObjectNotFoundError
     end
 
     # Initialize an empty model object and set its +resource+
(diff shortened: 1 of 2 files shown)


lib/active_fedora/core.rb
```diff
@@ -62,6 +62,13 @@ def reload
       refresh
       load_attached_files
       self
+    rescue Ldp::Gone => err
+      logger.info("Tried to reload an object that has been destroyed; " \
+                        "this is probably due to mismanagement of the object's " \
```
jcoyne: I would think this would most likely be the result of having two separate instances and destroying one of them and calling reload on the other.  I'm not sure that this conjecture is helpful.

no-reply (approved):

PR #1398, opened 2019-08-14
Support Rails 6

jcoyne:


no-reply:
is there interest in maintaining ActiveFedora 8.x?

if so, i wonder whether a fork is appropriate?

PR #1399, opened 2019-08-19
Releases version 13.0.0

jrgriffiniii:
https://github.com/samvera/active_fedora/pull/1385 introduced breaking changes with release 12.2.0.  This re-releases those updates with a major version.


no-reply:
Could we get in a fix for the generator conflicts, first?

jrgriffiniii:
Definitely, I'm working with that right now.

no-reply:
@jrgriffiniii i think i beat you to it https://github.com/samvera/active_fedora/pull/1400

PR #1400, opened 2019-08-19
Force overwrite of `solr/conf` files if they have changes

no-reply:
Blacklight generates Solr configuration into `solr/conf`. Since ActiveFedora now
generates its configurations into the same directory, any differing files will
be treated as conflicts by `Rails::Generators::Base`. Adding `force: true` to
the step that copies these files ensures that the AF versions of these files are
installed without requiring confusing user feedback at generation time.


no-reply:
prerequisite to #1399

PR #1403, opened 2019-09-10
Rails 6

no-reply:
Add support for Rails 6. Drop support for 5.1.


no-reply:
yeah, i was assuming we'd release 14.0.0 immediately.

no-reply:
new plan, it's made backwards compatible in #1407

PR #1406, opened 2019-09-13
Deprecate `#clear_changed_attributes`

no-reply:
This method can be replaced by `ActiveModel::Dirty#changes_applied`, which is
more resiliant to changing `ActiveModel` and `Rails` behavior.

However, since we call it in a callback, and downstream users may be relying on
that fact to trigger calls to their overrides, we can't avoid calling it
locally until a 14.0.0 release.

### DIFF
--- a/lib/active_fedora/aggregation/list_source.rb
+++ b/lib/active_fedora/aggregation/list_source.rb
@@ -15,6 +15,8 @@ module ActiveFedora
       # Overriding so that we don't track previously_changed, which was
       # rather expensive.
       def clear_changed_attributes
+        Deprecation.warn self.class, "#clear_changed_attributes is deprecated, use ActiveModel::Dirty#changes_applied instead."
+
         clear_changes_information
       end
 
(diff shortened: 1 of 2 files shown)


no-reply:
As noted, this method is hit by a callback. We have a few options:

  - Merge this and emit deprecation notices on every save until the user upgrades to 14.0.0 (this is the only breaking change in 14.0.0).
  - Introduce some metaprogramming to call a different method on callback, which would then emit a deprecation warning and route to the deprecated method *only* if it has been overridden.
  - Do nothing; never deprecate, release 14.0.0 with the [release notes](https://github.com/samvera/active_fedora/releases/tag/untagged-8eb8c8f845b5724e4a1d)

cjcolvar:
Option 1 seems like it might overflow logs in CI.
Option 2 seems better if it isn't too much work.
For option 3, do we know if anyone has actually overridden this method?  If not, then maybe the release notes are fine enough?

no-reply:
i think the proposed approach works, and may avoid the need to bump the major version #1405

PR #1407, opened 2019-09-13
Deprecate `#clear_changed_attributes`

no-reply:
This method can be replaced by `ActiveModel::Dirty#changes_applied`, which is
more resiliant to changing `ActiveModel` and `Rails` behavior.

However, since we call it in a callback, and downstream users may be relying on
that fact to trigger calls to their overrides, we continue calling it in the
case that it is both: not the base definition, and not the version overridden in
`ListSource`.

Callers to those two versions will see deprecation warnings, and users of the
callback will also see deprecation warnings. Their local behavior may be in
conflict with Rails 6.0.

### DIFF
--- a/lib/active_fedora/attributes.rb
+++ b/lib/active_fedora/attributes.rb
@@ -9,7 +9,30 @@ module ActiveFedora
       include Serializers
       include PrimaryKey
 
-      after_save :changes_applied
+      after_save :clear_changed_attributes_or_changes_applied
+      # @deprecated use #changes_applied instead
+      def clear_changed_attributes
+        Deprecation.warn ActiveFedora::Attributes, "#clear_changed_attributes is deprecated, use ActiveModel::Dirty#changes_applied instead."
+        @previously_changed = changes
+        clear_attribute_changes(changes.keys)
+      end
+
+      ##
+      # Call the deprecated method
+      #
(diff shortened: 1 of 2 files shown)


no-reply:
Note that this patch won't quite be backwards compatible for monkey patches on `ListSource`.

If you're monkey patching `ListSource`, you can jump in a lake.

lib/active_fedora/attributes.rb
```diff
@@ -9,7 +9,26 @@ module Attributes
...
+      after_save :clear_changed_attributes_or_changes_applied
+      def clear_changed_attributes
+        Deprecation.warn ActiveFedora::Attributes, "#clear_changed_attributes is deprecated, use ActiveModel::Dirty#changes_applied instead."
+        @previously_changed = changes
+        clear_attribute_changes(changes.keys)
+      end
+
+      ##
+      # Call the deprecated method
+      def clear_changed_attributes_or_changes_applied
```
cjcolvar: Is the plan to remove this method in AF 14?  If so, do we need a comment saying that it too is deprecated.

no-reply: maybe we should make it private. that seems like the best plan.

no-reply: marked private and deprecated.

PR #1418, opened 2020-01-24
Always rewind IO content even when the file is new.

jeremyf:
Co-authored-by: E. Lynette Rayle <elr37@cornell.edu>

### DIFF
--- a/spec/unit/file_spec.rb
+++ b/spec/unit/file_spec.rb
@@ -192,6 +192,23 @@ describe ActiveFedora::File do
     end
   end
 
+  context 'when file is new and content behaves like io' do
+    let(:file_content) { "hello world" }
+
+    before do
+      af_file.uri = "http://localhost:8983/fedora/rest/test/1234/abcd"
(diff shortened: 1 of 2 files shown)


spec/unit/file_spec.rb
```diff
@@ -192,6 +192,23 @@
...
+    before do
+      af_file.uri = "http://localhost:8983/fedora/rest/test/1234/abcd"
+      af_file.content = StringIO.new(file_content)
+      allow(af_file).to receive(:new_record?).and_return(true)
+    end
+
+    describe "#content" do
+      it 'can be re-read' do
+        expect(af_file.content.read).to eq file_content
+        expect(af_file.content.read).to eq file_content
```
straleyb: I want to ensure that "re-read" ACTUALLY means you can call #read 2 times on the file's content. If thats the case then PR looks good

jeremyf: I think that's a reasonable clarification on the spec. As this PR has now been merged to 3 other branches, I'm going to resolve the comment.

PR #1421, opened 2020-06-09
Updating the EBUCore term for `RDF::Vocab::EBUCore.filename`

jrgriffiniii:
This is necessary given that this term was removed: https://github.com/ruby-rdf/rdf-vocab/blob/develop/lib/rdf/vocab/ebucore.rb#L7117. Please see https://github.com/samvera/hydra-derivatives/pull/221 for context.

### DIFF
--- a/active-fedora.gemspec
+++ b/active-fedora.gemspec
@@ -14,30 +14,29 @@ Gem::Specification.new do |s|
   s.license = "Apache-2.0"
   s.required_ruby_version = '~> 2.4'
 
-  s.add_dependency 'rsolr', '>= 1.1.2', '< 3'
-  s.add_dependency "activesupport", '>= 5.2'
   s.add_dependency "activemodel", '>= 5.2'
+  s.add_dependency "activesupport", '>= 5.2'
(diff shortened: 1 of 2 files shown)


jrgriffiniii:
I am not certain why the CircleCI status for my fork is not displaying here: https://app.circleci.com/pipelines/github/jrgriffiniii/active_fedora/49/workflows/5af86cf0-f082-4e00-aba8-ecd23944a423

I will attempt to update the configuration and force-push to my branch.

no-reply:
this needs backports to 11.x, and other supported AF versions

no-reply:
looking at this closer, i think we should revert it.

changing this predicate is going to require users with existing data to migrate all existing objects.

jrgriffiniii:
A new clone of hyrax `3.0.0-rc1` installs `3.0.14` of `rdf-vocab` for me, and this release still offers the `filename` property.

no-reply:
it looks to me like the change was first released in `v3.1.5` https://github.com/ruby-rdf/rdf-vocab/commit/d07779dbcc4cca5c34a96d6118c4aa8ee23673a9

PR #1422, opened 2020-06-11
Updating the EBUCore term for filename properties (backporting into 11.5.z)

jrgriffiniii:
Backporting this to the 11.5.z release series.


no-reply:
closing this. we need to backport #1423 instead.

PR #1424, opened 2020-06-12
remove `StandardFilterFactory`

no-reply:
`StandardFilterFactory` hasn't done anything since Solr 3. Remove it for better
forward support.


no-reply:
sorry about the whitespace changes. i can isolate those, if we'd prefer, but it's a pain (my editor is configured to remove trailing whitespace) and it seems good to remove it anyway, since this gets injected into applications.

rotated8:
As an aside, in GitHub, you can hide whitespace changes by adding `?w=1` to the end of the url on the files tab, e.g. https://github.com/samvera/active_fedora/pull/1424/files?w=1

PR #1430, opened 2020-06-22
Rewording constants for greater accessibility

jeremyf:
Related to https://github.com/samvera/hyrax/commit/ad5e64590f91a8fdf2c3ff812b278ba73eee487e

### DIFF
--- a/lib/active_fedora/attribute_methods.rb
+++ b/lib/active_fedora/attribute_methods.rb
@@ -13,7 +13,7 @@ module ActiveFedora
       end
     end
 
-    BLACKLISTED_CLASS_METHODS = %w(private public protected allocate new name parent superclass).freeze
+    DANGEROUS_CLASS_METHODS = %w(private public protected allocate new name parent superclass).freeze
 
     class GeneratedAttributeMethods < Module; end # :nodoc:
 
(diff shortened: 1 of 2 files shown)


cjcolvar:
I'm :+1: to this change.  Do you think these constants are part of the public API that would require a major version bump?  If so, then we probably need deprecation warnings.

no-reply:
looks like this needs a rebase.

we could mitigate the major version situation by aliasing the constants for the time being?

lib/active_fedora/relation/delegation.rb
```diff
@@ -7,7 +7,7 @@ module Delegation # :nodoc:
     # may vary depending on the klass of a relation, so we create a subclass of Relation
     # for each different klass, and the delegations are compiled into that subclass only.
 
-    BLACKLISTED_ARRAY_METHODS = [
```
jcoyne, line 10: As this was cloned from ActiveRecord, this might be a better change: https://github.com/rails/rails/commit/9d79334a1dee67e31222c790e231772deafcaeb8

lib/active_fedora/attribute_methods.rb
```diff
@@ -13,7 +13,7 @@ def self.set_name_cache(name, value)
       end
     end
 
-    BLACKLISTED_CLASS_METHODS = %w(private public protected allocate new name parent superclass).freeze
+    DANGEROUS_CLASS_METHODS = %w(private public protected allocate new name parent superclass).freeze
```
jcoyne, line 16: As this was copied from ActiveRecord, maybe follow the constant they have chosen: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/attribute_methods.rb#L26

cjcolvar:
Closing since this was superseded by and resolved in #1501

PR #1431, opened 2020-06-27
prepare release 13.1.3

no-reply:
Release 13.1.3

We need a release with the `rdf-vocab` pin.


no-reply:
i'm unsure if there's action i need to take here re the new changelog generator. is that documented somewhere?

PR #1436, opened 2020-10-23
Own the legacy EBUCore property instead of pinning all of rdf-vocab

cbeer:


no-reply (approved):

lib/active_fedora/with_metadata/default_schema.rb
```diff
@@ -3,8 +3,19 @@
...
+          domain: "ebucore:Resource".freeze,
+          label: "File name".freeze,
+          range: "xsd:string".freeze,
+          type: "rdf:Property".freeze
+      end
+    end
+
     property :label, predicate: ::RDF::RDFS.label
-    property :file_name, predicate: ::RDF::Vocab::EBUCore.filename
+    property :file_name, predicate: legacy_ebucore_filename_property.filename
```
no-reply, line 18:
```suggestion
    property :file_name, predicate: legacy_ebucore_vocabulary.filename
```

PR #1438, opened 2020-10-28
Allow initializing OrderedList with an existing NodeCache

aploshay:
The initialization line:
```ruby
        @node_cache ||= NodeCache.new
```
suggests that the node cache might have been supplied through other means, but there's currently not possible; the proposed change allows for it.

This is a change we made within a project to fix a production bug in the component that built an arbitrarily large and deep set of nested OrderedList objects, before saving it -- without the ability to pass along an existing, paritlally-populated, node cache, we were getting collisions on ids once the whole set was persisted.

### DIFF
--- a/lib/active_fedora/orders/ordered_list.rb
+++ b/lib/active_fedora/orders/ordered_list.rb
@@ -12,7 +12,7 @@ module ActiveFedora
       #   stored.
       # @param [::RDF::URI] head_subject URI of head node in list.
       # @param [::RDF::URI] tail_subject URI of tail node in list.
-      def initialize(graph, head_subject, tail_subject, node_cache = NodeCache.new)
+      def initialize(graph, head_subject, tail_subject, node_cache: NodeCache.new)
         @graph = if graph.respond_to?(:graph)
                    graph.graph.data
                  else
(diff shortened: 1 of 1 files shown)


lib/active_fedora/orders/ordered_list.rb
```diff
@@ -12,15 +12,15 @@ class OrderedList
       #   stored.
       # @param [::RDF::URI] head_subject URI of head node in list.
       # @param [::RDF::URI] tail_subject URI of tail node in list.
-      def initialize(graph, head_subject, tail_subject)
+      def initialize(graph, head_subject, tail_subject, node_cache = NodeCache.new)
```
no-reply, line 15: seems like a straightforward improvement to me. could i suggest we use a named argument, instead? this should be more tolerant toward any potential extensions (since it doesn't depend on argument order).

```suggestion
      def initialize(graph, head_subject, tail_subject, node_cache: NodeCache.new)
```

aploshay: Sure thing, change committed

no-reply (approved):

PR #1439, opened 2020-11-12
add `omitNorms="true"` to solr `dateRange` as fix for Solr 8.6.3

no-reply:
this is (maybe?) a fix for https://github.com/samvera/hyrax/pull/4542.


cjcolvar (approved):
:1st_place_medal: Thanks so much for figuring this out!
This allowed solr to startup correctly for me with solr_wrapper and a docker container using the collection creation process in hyrax's docker-compose.

PR #1442, opened 2021-03-03
Short circuit ActiveModel::Dirty for nodes_will_change!

cjcolvar:
ActiveModel::Dirty will attempt to load the entire ordered list when calling nodes_will_change! which leads to a stack level too deep exception in some cases (samvera/hyrax#4581).

This approach was also taken in ActiveFedora::File: 7c8bbbe#diff-28356c4daa0d55cbaf97e4269869f510R100-R103


no-reply:
should this go to `master` first?

no-reply:
this was already removed/fixed in AF 13+. merging it for release in 12.x

PR #1445, opened 2021-05-14
compact some potentially large scale iterations in reindex_all

no-reply:
only loop through possibly large collections once.

in the case of the repository root, the number of object in these collections is
potentially hundreds of thousands or millions of objects. restricting ourselves
to a single iteration is helpful, and `RDF::Enumerable#each_object` presents a
single upstream method to house memory optimizations.

the older implementation seemed pre-occupied with preventing repeated `#to_s`
calls. this isn't necessary since these are cheap variable dereferences in
almost all relevant cases.


cjcolvar (approved):
Looks great!

PR #1457, opened 2021-11-05
Fedora 6 support

cjcolvar:
This is a proof of concept that builds upon work done by @tpendragon and @stkenny.  It does not consider backwards-compatibility yet and naively tries to get tests passing without consideration for realigning code to completely match Fedora's API.


cjcolvar:
Closed via #1504

PR #1464, opened 2022-01-11
Don't track changes or attempt to persist server-managed properties on file metadata nodes

cjcolvar:
Upstream port of monkey-patch in Hyrax required for wings (https://github.com/samvera/hyrax/pull/5320)

This PR introduces an attribute on properties in the metadata schemas called `:server_managed`.  When this attribute is set to true on a property it allows a value to be set via the generated setter but the change isn't tracked by `ActiveModel::Dirty` and subsequently isn't sent to the server as part of the changed values when saving.

### DIFF
--- a/lib/active_fedora/with_metadata/default_schema.rb
+++ b/lib/active_fedora/with_metadata/default_schema.rb
@@ -21,6 +21,6 @@ module ActiveFedora::WithMetadata
     property :date_modified, predicate: ::RDF::Vocab::EBUCore.dateModified
     property :byte_order, predicate: SweetJPLTerms.byteOrder
     # This is a server-managed predicate which means Fedora does not let us change it.
-    property :file_hash, predicate: ::RDF::Vocab::PREMIS.hasMessageDigest
+    property :file_hash, predicate: ::RDF::Vocab::PREMIS.hasMessageDigest, server_managed: true
   end
 end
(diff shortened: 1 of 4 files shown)


lib/active_fedora/with_metadata/default_schema.rb
```diff
@@ -21,6 +21,6 @@ def self.legacy_ebucore_vocabulary
     property :date_modified, predicate: ::RDF::Vocab::EBUCore.dateModified
     property :byte_order, predicate: SweetJPLTerms.byteOrder
     # This is a server-managed predicate which means Fedora does not let us change it.
-    property :file_hash, predicate: ::RDF::Vocab::PREMIS.hasMessageDigest
+    property :file_hash, predicate: ::RDF::Vocab::PREMIS.hasMessageDigest, server_managed: true
```
no-reply, line 24: i'm really into this pattern. 💯

lib/active_fedora/with_metadata/metadata_node.rb
```diff
@@ -70,6 +70,10 @@ def changes_for_update
           ChangeSet.new(self, self, changed_attributes.keys).changes
         end
 
+        def server_managed_properties
+          @server_managed_properties ||= properties.select { |k,v| v[:server_managed] }.symbolize_keys.keys
```
no-reply, line 74: this is minor, but i think this will have notably fewer allocations:

```suggestion
          @server_managed_properties ||= properties.select { |k,v| v[:server_managed] }.keys.map(&:to_sym)
```

no-reply (approved):
thanks for picking this up and backporting!

this pattern is a lot nicer even than the hack in Wings. 🥇

Issue #497, opened 2014-10-22
Do PUT requests with the exact modified date (precise to a thousandth of a second)


jcoyne:
On edge fcrepo4, we get a 409 Conflict error when we attempt to do an update:

```
Could not remove triple containing predicate http://fedora.info/definitions/v4/repository#lastModified to node /test/9f/30/63/86/9f306386-f98f-4fc0-af2f-8a85cb2257ad
```

I believe this is because we are sending a lastModifed value with whole second precision, but Fedora has the value precise to a thousandth of a second.   We are currently getting this value from the header, but we might need to get it from the graph in order to get the correct precision.

PR #1468, opened 2022-02-23
Conditionally apply monkey patch to handle versions of RDF 3.2.5+

### LINKED ISSUE #1466: rdf 3.2.5 causes superclass mismatch for class DateTime
` .../active-fedora-13.2.5/lib/active_fedora.rb:16:in '<class:Literal>': superclass mismatch for class DateTime (TypeError)`

rdf 3.2.4, no problem

cjcolvar:
Resolves #1466 

RDF 3.2.5 changed the superclass of DateTime which broke the monkey patch in ActiveFedora.  This PR conditionally applies the monkey patch if the class exists.  

It appears that the monkey patch might not be necessary with RDF 3.2.5 due to the changes to the class in RDF (see https://github.com/ruby-rdf/rdf/releases/tag/3.2.5) so I didn't attempt to create a new version of the monkey patch.  But I removed the monkey patch and all the tests passed locally with RDF 3.2.4 so maybe the changes aren't actually being tested properly or aren't necessary anymore due to improvements in Fedora 4 or ActiveFedora since the monkey patch was applied in 2014? :man_shrugging: 

Within the CI environment the monkey patch should still be getting applied for the ruby 2.5 build because RDF 3.2.5 requires 2.6+.

### DIFF
--- a/lib/active_fedora.rb
+++ b/lib/active_fedora.rb
@@ -11,21 +11,26 @@ require 'active_triples'
 # Monkey patching RDF::Literal::DateTime to support fractional seconds.
 # See https://github.com/samvera/active_fedora/issues/497
 # Also monkey patches in a fix for timezones to be stored properly.
-module RDF
-  class Literal
-    class DateTime < Literal
-      ALTERNATIVE_FORMAT   = '%Y-%m-%dT%H:%M:%S'.freeze
-      DOT                  = '.'.freeze
-      EMPTY                = ''.freeze
-      TIMEZONE_FORMAT      = '%:z'.freeze
+#
+# RDF 3.2.5 changes the superclass of DateTime to RDF::Temporal
+# TODO: Determine if this monkey-patch is needed even with RDF pre-3.2.5 (All the tests pass without it)
+if RDF::Literal::DateTime.superclass == RDF::Literal
+  module RDF
+    class Literal
+      class DateTime < Literal
+	ALTERNATIVE_FORMAT   = '%Y-%m-%dT%H:%M:%S'.freeze
+	DOT                  = '.'.freeze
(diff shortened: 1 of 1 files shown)


lib/active_fedora.rb
```diff
@@ -11,21 +11,26 @@
...
-        @string ||= begin
-          # Show nanoseconds but remove trailing zeros
-          nano = @object.strftime('%N').sub(/0+\Z/, EMPTY)
-          nano = DOT + nano unless nano.blank?
-          @object.strftime(ALTERNATIVE_FORMAT) + nano + @object.strftime(TIMEZONE_FORMAT)
-        end
+#
+# RDF 3.2.5 changes the superclass of DateTime to RDF::Temporal
+# TODO: Determine if this monkey-patch is needed even with RDF pre-3.2.5 (All the tests pass without it)
+if RDF::Literal::DateTime.superclass == RDF::Literal
```
jrgriffiniii, line 17: 💯

jrgriffiniii (approved):

elrayle:
It looks like the intention of the monkey patch is to add precision to the date for put requests to Fedora.  (See Issue #497.)

* does AF still support Fedora 4
* do later versions of Fedora not require the same level of precision

cjcolvar:
I believe ActiveFedora only supports Fedora 4 at the moment.  I wonder if this issue was encountered during an early version of Fedora 4 and fixed in a subsequent version.  The tests including integration tests pass against Fedora 4 so it doesn't seem like #497 is still a problem.  The change in rdf 3.2.5 changing the underlying implementation from `::Date` to `::DateTime` I think should also have brought in the increased precision.

PR #1476, opened 2022-03-09
Fix obsolete URI.encode call

jrgriffiniii:
Derived from rebasing https://github.com/samvera/active_fedora/pull/1461


jrgriffiniii:
This is now rebased and passing on CircleCI. Thank you again for the actual updates @apertome!

cjcolvar:
When making the 12.2.4 release to fix the rdf issue I also backported the fix for this from main (https://github.com/samvera/active_fedora/pull/1474).  So I'm not sure if this PR makes sense to do anymore.  The approach taken in main is different and looking at the two I would tend to favor the one in main since it doesn't bring in any new dependencies.

PR #1478, opened 2022-04-18
Updating the CircleCI configuration to support nightly builds

### LINKED ISSUE #1477: Configure nightly CircleCI test suite executions
Derived from https://github.com/samvera/maintenance/issues/89

jrgriffiniii:
Resolves #1477 and addresses the following:

- Updating the configuration to test against Ruby 2.7, 2.6, and 2.5 releases
- Updating the configuration to test against Rails 6.0.4.7, 5.2.4
- Setting the default Rails release to be used as the latest 6.0 release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  samvera: samvera/circleci-orb@0
+  samvera: samvera/circleci-orb@1.0
 jobs:
   bundle_lint_test:
     parameters:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -47,31 +47,79 @@ jobs:
...
+                - main
+    jobs:
+      # Ruby 2.7 releases
       - bundle_lint_test:
-          name: ruby2-4_rails5-2
-          ruby_version: 2.4.9
-          rails_version: 5.2.4
+          name: ruby2-7_rails6-1
+          ruby_version: 2.7.5
+          rails_version: 6.1.5
```
cjcolvar, line 91: I'm not seeing rails 6.1 being tested in the commit-triggered build.  How about adding it to the on-commit build matrix?  If it doesn't work then I would think it should be removed from both on-commit and nightly builds until it can be resolved.

cjcolvar:
Resolved in #1483

PR #1493, opened 2023-02-08
Avoid modifying frozen array

cjcolvar:
When attempting to upgrade Avalon to rails 7.0, I ran into an error about modification of frozen arrays, but the real issue was that Rails 6.1 included a major change in `ActiveModel::Error` turning it into an array of objects instead of a simple hash.  The API changed and I had to find an approach that worked for all versions of Rails that ActiveFedora supports.  I think the approach I took is a stop gap and that we'll want to rework this using the new API more and possibly rethinking this section entirely.  Currently the errors are surfaced on the parent object by composing the child association name with the attribute of the child which has the error.  This is problematic because the new `ActiveModel::Error` implementation checks if error keys are actual attributes on the model object.  For example, ```errors.messages_for("members.foo")` raises `NoMethodError Exception: undefined method `members.foo'```.

It appears that there weren't any tests for this part of the code so it slipped by the prior work on supporting rails 7.

### DIFF
--- a/lib/active_fedora/autosave_association.rb
+++ b/lib/active_fedora/autosave_association.rb
@@ -243,10 +243,9 @@ module ActiveFedora
 
         unless valid = record.valid?
           if reflection.options[:autosave]
-            record.errors.each do |attribute, message|
+            record.errors.messages.each do |attribute, messages|
               attribute = "#{reflection.name}.#{attribute}"
-              errors[attribute] << message
-              errors[attribute].uniq!
+              messages.map { |message| errors.add(attribute, message: message) }
             end
           else
             errors.add(reflection.name)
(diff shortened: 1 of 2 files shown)


lib/active_fedora/autosave_association.rb
```diff
@@ -243,10 +243,9 @@ def association_valid?(reflection, record)
 
         unless valid = record.valid?
           if reflection.options[:autosave]
-            record.errors.each do |attribute, message|
+            record.errors.messages.each do |attribute, messages|
```
cjcolvar, line 246: `#messages` returns a hash with attributes as keys and arrays of messages as values in rails 5.2 through 7.0.  This was the only way I could find that worked in all supported rails versions.

jeremyf (approved):

PR #1511, opened 2026-09-22
Allow RSolr 3

maxkadel:
RSolr 3 is a breaking change because it drops Faraday 0.x support, which this gem already does not support. (`s.add_dependency 'faraday', '>= 1.0'`)

RSolr 3 also allows for putting Solr basic authentication in headers, rather than in the Solr url itself. Not trying to implement that at this time, but want it to be an option downstream.


cjcolvar:
Thanks @maxkadel for the PR!  

I'm guessing that you may want to target the `15.x-stable` branch as it is the last of the Fedora 4.x line.  `main` might work with Fedora 4.x but is meant to work with Fedora 6/7.  I'm not certain that switching the branch will clear up the CI failures but it should hopefully resolve the out-of-date with base branch warning.

maxkadel:
> Thanks @maxkadel for the PR!
> 
> I'm guessing that you may want to target the `15.x-stable` branch as it is the last of the Fedora 4.x line. `main` might work with Fedora 4.x but is meant to work with Fedora 6/7. I'm not certain that switching the branch will clear up the CI failures but it should hopefully resolve the out-of-date with base branch warning.

ah, yes, I'll close this one and re-open against the right one, sorry about that!

PR #1512, opened 2026-09-22
Allow RSolr 3

maxkadel:
RSolr 3 is a breaking change because it drops Faraday 0.x support, which this gem already does not support. (`s.add_dependency 'faraday', '>= 1.0'`)

RSolr 3 also allows for putting Solr basic authentication in headers, rather than in the Solr url itself. Not trying to implement that at this time, but want it to be an option downstream.


cjcolvar:
@maxkadel I got CI green in main (#1513) and I think if you pull in #1508 and #1513 then hopefully this PR will go green too. :crossed_fingers:

PR #1513, opened 2026-09-22
Fix CI and test with latest ruby / rails versions

cjcolvar:


spec/unit/forbidden_attributes_protection_spec.rb
```diff
@@ -21,6 +21,7 @@ def dup
           duplicate.instance_variable_set :@permitted, @permitted
         end
       end
+      alias stringify_keys dup
```
maxkadel: Can you help me understand why this belongs in the spec, and not in the code?

maxkadel (approved):
Approving since I don't want to block, ~but I would like you to double check the location of the stringify_keys alias before merging.~ (get it now)

Does this force a breaking version, since it drops Rails 6 tests?

.circleci/config.yml
```diff
@@ -86,54 +87,45 @@ jobs:
...
-          name: ruby3-2_rails7-0
-          ruby_version: 3.2.5
-          rails_version: 7.0.8.4
+          name: ruby3-4_rails8-0
+          ruby_version: 3.4.10
+          rails_version: 8.0.5.1
       - bundle_lint_test:
-          name: ruby3-2_rails6-1
-          ruby_version: 3.2.5
-          rails_version: 6.1.7.8
```
maxkadel: If we're no longer testing against Rails 6, do we need to change the Gemspec and bump a major version?

spec/unit/forbidden_attributes_protection_spec.rb
maxkadel: Ok, I see, just fixing the mock.

.circleci/config.yml
cjcolvar: The approach I've taken before is to not technically disallow old versions but drop testing because it can be too burdensome to keep CI working with old versions.  It might still work with older versions of rails so I don't want to block users from trying but your mileage may vary.  Does that seem okay to you?

maxkadel: Yes, I think that makes sense, thanks for explaining

